Issue #5, opened 2014-12-13
Put support


grobie:
Hey,

great library! Looks very promising. I can see there is `CreateEmptyFile` method, but I can't see any other methods to write files to hdfs. Are there plans to add support for that, also for  put/copyFromLocal/moveFromLocal/etc. CLI commands?

Sorry if I missed the note about that.

Thanks

mindscratch:
very interested in this as well, I appreciate the work.

colinmarc:
@mindscratch thanks for the bump =) If you're interested on living on the edge, I think the tests are passing in the PR I referenced above (#12). It doesn't support failing over if a datanode dies, though.

andrewlee302:
Hey, are you progressing on write supports? I am taking a HDFS-based distributed computing framework in golang, so i am eager to using your writing function instead of webREST interfaces.

colinmarc:
Hey Andrew,

I Haven't made any progress recently, but if you check out the linked
branch the interfaces are there, and they work. The missing bits are
failover and comprehensive testing, but you could still develop against the
branch.

Sorry about the trouble!

imoverclocked:
Hey @colinmarc,

  Thanks for the library and cli tools! It feels blazingly fast compared to using the standard hdfs cli or loading a jvm in general :)

  I see you wrote put support and I would like to use it. Since there is no support for datanode failover, what happens when a node does fail? Is there any kind of panic/error returned or do we currently just keep going like nothing happened?

Cheers,
-Tim

colinmarc:
@imoverclocked If the test failures are any indication, I think you'll get some (byzantine) errors. Hopefully not a panic.

colinmarc:
Closed in https://github.com/colinmarc/hdfs/pull/12

PR #12, opened 2015-03-03
[wip] Write support

colinmarc:
For #5. The protocol is pretty involved!


noahhl:
Feels a little weird to be commenting on an incomplete pull request, but I'll give it a shot :)

We've been using this write support in various ways for a few months, and it's great (thanks!). I finally noticed today that it wasn't respecting replication settings, either the cluster defaults or specified via `client.CreateFile`. Easy enough to fix, a la:

```diff
diff --git a/file_writer.go b/file_writer.go                                                                                                                                                                                                                                                                         [0/5990]
index bd5d958..4c3cf81 100644
--- a/file_writer.go
+++ b/file_writer.go
@@ -56,7 +56,7 @@ func (c *Client) CreateFile(name string, replication int, blockSize int64, perm
                ClientName:   proto.String(rpc.ClientName),
                CreateFlag:   proto.Uint32(1),
                CreateParent: proto.Bool(false),
-               Replication:  proto.Uint32(uint32(1)),
+               Replication:  proto.Uint32(uint32(replication)),
                BlockSize:    proto.Uint64(uint64(blockSize)),
```

Once that was straightened out though, I started getting intermittent `Invalid ack sequence number` errors upon writing to files with replication > 1. After some digging, it seems to always be the case that the returned sequence number in the acknowledgement is one greater than the first packet in the `blockWriteStream` (e.g., if my understanding is correct, the datanode is one sequence number ahead of the client). 

I'm not sure if it's a thread safety issue or acknowledgements coming out of order from different datanodes in the pipeline. I've been poking at it for a while and will keep doing so, but figured I'd see if you had an obvious thought.

noahhl:
The replication factor thing might have actually been a red herring or just an exacerbating factor -- I think it's thread safety after all. Locking `packets` with a mutex in both `flush()` and `ackPackets()` seems to do the trick; I can consistently tickle ack-out-of-sequence problems without the mutex, and not at all with.

colinmarc:
@noahhl Ahh, thanks for pointing that out. I just got back from vacation, and I'll need to page this all back into memory, but I'll definitely take a look. Pretty plausible that there are thread safety issues =)

andrewlee302:
Hey, i appreciate your works and the write-support branch much. I'm using your library as my  distributed computing framework for half a year. I've noticed the I/O throughput differences from Hadoop. But i'm not sure it's other factors that affected I/O and i know less about the implementation than you, of course. ^_^

1. So i'm curious to know the possible performance overheads or benifits based your code implementation versus Hadoop read or write. 
2. I'd also like to know the limitations about the library on common hdfs usage. 
3. Is that possble that jvm's overheads are more than go if only I/O opereations are taken into account.

colinmarc:
Hi @andrewlee302! Thanks for writing in.

The branch hasn't been merged yet because I haven't sat down and finished up the failover stuff. In the HDFS protocol, the client handles all of the failover steps, including setting up the new replication pipeline and instructing datanodes to replicate blocks, so it's very complicated.

I'd be really interested in seeing the benchmarks you're talking about! I didn't focus much on performance yet, but I wouldn't be surprised if it was a bit better or a bit worse. One major difference that comes to mind is that the java client, for some reason, buffers an entire block to disk before copying it up to the datanode. In this library, I only do in-memory buffering.

andrewlee302:
Hi @colinmarc ! I'd like to show some experimental data to you.
The testbed is a cluster of 8 nodes interconnected by a 1 Gigabit Ethernet switch, 1 node for hdfs master and 7 other nodes as slaves, each of which is equipped with two Intel Xeon E5620 CPU processors (2.4 GHz). *HDFS configuration: block size: 256M, replication: 3* .
GoDataMPI uses your library to do only read or write operations like the TestDFSIO in Hadoop benchmarks. There are two experiments, one is for writing 84G data(3G* 28 files) and the other one is for reading the former data. Hadoop uses 28 slots (7 slaves * 4 ), and GoDataMPI uses 28 processes either, in other words, one process opereate just one file(3G).

P.S. There is litter data in shuffle stage, so network traffic is mainly on hdfs operations.

## 84G Writing: disk-io bandwidth and net-io bandwidth.
84g-write-disk_io
80g-write-net_usage

## 84G Reading: disk-io bandwidth and net-io bandwidth.
84g-read-disk_io
80g-read-net_usage

## Statistics
*Item* | *Throughput(MB/Sec)* | *Average IO rate(MB/Sec)* | *Execution time(s)*
-------- | ------ | ----- | ----
84G- *Write-Hadoop* | 8.4599 | 8.5480 | 441 |
84G- *Write-GoDataMPI* | 8.609 | 8.250 | 405 |
84G- *Read-Hadoop* | 11.075 | 11.879 | 378 |
84G- *Read-GoDataMPI* | 16.691 | 19.250 | 259 |

## Conclusion
The writing performance is almost the same. But GoDataMPI performs better than Hadoop on reading.
In addition, its execution time is less than Hadoop because of the jvm start-up overheads.

## Questions
I'm confused by your words about the replications. Go client doesn't replicate the data instead of jvm doing 3 replications when writing 84G data, does it? 
1. If it doesn't do, go should execute writing much faster than Hadoop because of replication overheads, and execute reading much slower than Hadoop because of the locality.
2. If it indeed do, the similar performance on writing makes sense.  But why the reading performances are so different?

Thanks for reviewing. I'm looking forward to your comments !!!

colinmarc:
@andrewlee302 thanks, this is very interesting! In response to your questions:

> If it doesn't do, go should execute writing much faster than Hadoop because of replication overheads, and execute reading much slower than Hadoop because of the locality.

The go client does replicate correctly - it just doesn't set up a new replication pipeline if one of the datanodes in the pipeline dies during a read. That's the only thing missing from this PR.

> If it indeed do, the similar performance on writing makes sense. But why the reading performances are so different?

Great question! I don't really know. Go is a very "fast" language, in the sense that it is easy to write performant code in it. And it's possibly just generally faster than the JVM for i/o stuff. My guess, though, is that the java client is just doing something dumb that no one has caught. One thing you see a lot in java code is allocating new buffers instead of reusing them, just because java doesn't really give you good tools to reuse memory. So something like that. I'm not about to go profile the java client to find out =).

One thing I'm curious about would be the performance if checksumming is disabled. I know that snakebite turns it off by default, for example. Would you be willing to run a test with that? For the purposes of the test, you should be able to just comment out this: https://github.com/colinmarc/hdfs/blob/master/rpc/block_read_stream.go#L125-L132

wangkuiyi:
Report a write error:

When I run the following small program
```
func main() {                                                                                                                                                                                                                                                                                                                                                             
        namenode := flag.String("namenode", "localhost:9000", "Address of HDFS namenode")                                                                                                                                                                                                                                                                                 
        flag.Parse()                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                          
        dfs, e := hdfs.New(*namenode)                                                                                                                                                                                                                                                                                                                                     
        if e != nil {                                                                                                                                                                                                                                                                                                                                                     
                log.Fatal(e)                                                                                                                                                                                                                                                                                                                                              
        }                                                                                                                                                                                                                                                                                                                                                                 
        w, e := dfs.Create("/hello.txt")                                                                                                                                                                                                                                                                                                                                  
        if e != nil {                                                                                                                                                                                                                                                                                                                                                     
                log.Fatal(e)                                                                                                                                                                                                                                                                                                                                              
        }                                                                                                                                                                                                                                                                                                                                                                 
        r, e := w.Write([]byte("Hello, World!\n"))                                                                                                                                                                                                                                                                                                                        
        if e != nil {                                                                                                                                                                                                                                                                                                                                                     
                log.Fatal(e)  ///////////// Fatal here.                                                                                                                                                                                                                                                                                                                                            
        }                                                                                                                                                                                                                                                                                                                                                                 
        fmt.Println("written ", r)                                                                                                                                                                                                                                                                                                                                        
        w.Close()                                                                                                                                                                                                                                                                                                                                                         
}    
```
The line of `w.Write(...)` returns an error:
```
$ go run hdfs.go
2015/12/01 13:23:16 create /hello.txt: addBlock call failed with ERROR_APPLICATION (java.io.IOException)
exit status 1
```

colinmarc:
Since it seems like a bunch of people are using this already, I'm going to merge it without the ability to orchestrate failovers. I did some cleanup, and added `hdfs put`.

PR #14, opened 2015-06-18
 Fix race condition on blockWriteStream

yjh0502:
Fix race condition on blockWriteStream.packets. Use channel instead to
 slice to check ack messages. It fixes race condition on packets and
 gives proper rate controls with channel size.

### DIFF
--- a/rpc/block_write_stream.go
+++ b/rpc/block_write_stream.go
@@ -3,11 +3,11 @@ package rpc
 import (
 	"bufio"
 	"bytes"
-	"github.com/golang/protobuf/proto"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	hdfs "github.com/colinmarc/hdfs/protocol/hadoop_hdfs"
+	"github.com/golang/protobuf/proto"
 	"hash/crc32"
 	"io"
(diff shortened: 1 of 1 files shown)


rpc/block_write_stream.go
```diff
@@ -115,8 +118,6 @@ func (s *blockWriteStream) finish() error {
 	if err != nil {
 		return err
 	}
-
-	// The last packet has no data; it's just a marker that the block is finished.
```
colinmarc: Any reason for removing this comment?

colinmarc:
Hey! Thanks for the contribution!

This looks good - can you also add a test in another PR that fails without this code? If it ends up being too gnarly, that's fine.

rpc/block_write_stream.go
yjh0502: The comment is accidentally deleted, I just submitted a fix.

yjh0502:
Thanks for a review. The problem only appears on highly loaded system, so I'm not sure if I can write a testcase that reproduces the problem.

colinmarc:
Ok, that makes sense. Thanks again!

PR #15, opened 2015-06-23
MkdirAll: Do not return error when already exists

yjh0502:
Documentation states that MkdirAll does not return error when directory
already exists, however is not. Fix MkdirAll to return nil when
directory already exists.

### DIFF
--- a/mkdir_test.go
+++ b/mkdir_test.go
@@ -64,6 +64,18 @@ func TestMkdirAll(t *testing.T) {
 	assert.EqualValues(t, 0, fi.Size())
 }
 
+func TestMkdirAllExists(t *testing.T) {
+	client := getClient(t)
+
+	baleet(t, "/_test/dir4")
+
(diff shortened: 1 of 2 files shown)


colinmarc:
Hey, thanks so much! Can you reopen this against master? I'm going to close this one.

mkdir_test.go
```diff
@@ -64,6 +64,18 @@ func TestMkdirAll(t *testing.T) {
...
 
+func TestMkdirAllExists(t *testing.T) {
+	client := getClient(t)
+
+	baleet(t, "/_test/dir4")
+
+	err := client.MkdirAll("/_test/dir4/foo", mode)
+	require.NoError(t, err)
+
+	err := client.MkdirAll("/_test/dir4/foo", mode)
```
colinmarc: This should be `err =`, not `err :=`.

PR #16, opened 2015-06-23
MkdirAll: Do not return error when already exists

yjh0502:
Documentation states that MkdirAll does not return error when directory
already exists, however is not. Fix MkdirAll to return nil when
directory already exists.


colinmarc:
Woohoo! Thank you =)

PR #17, opened 2015-06-26
fetch getopt from github instead of googlecode

yjh0502:
`getopt` package is moved to github, so there's no reason to use old
code.google.com link. It also removes mercurial dependency on installing
`hdfs` command.


colinmarc:
looks great, thanks!

PR #18, opened 2015-10-02
Use environment for determining user per https://github.com/golang/go/issues/6376

ebartels:
I encountered https://github.com/golang/go/issues/6376 and this also allows one to set `env USER=$SOMEOTHERUSER hdfs`

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,8 +20,15 @@ type Client struct {
 // New returns a connected Client, or an error if it can't connect. The user
 // will be the user the code is running under.
 func New(address string) (*Client, error) {
-	username := os.Getenv("USER")
-	return NewForUser(address, username)
+	username := os.Getenv("HADOOP_USER_NAME")
+	if username != "" {
+		return NewForUser(address, username)
+	}
+	currentUser, err := user.Current()
+	if err != nil {
+		return nil, err
+	}
+	return NewForUser(address, currentUser.Username)
 }
 
 // NewForUser returns a connected Client with the user specified, or an error if
--- a/client.go
+++ b/client.go
@@ -17,18 +17,28 @@ type Client struct {
 	defaults *hdfs.FsServerDefaultsProto
 }
 
-// New returns a connected Client, or an error if it can't connect. The user
-// will be the user the code is running under.
-func New(address string) (*Client, error) {
+// Username returns the HADOOP_USER_NAME environment supplied user if provided
+// or the operating system current user
+func Username() (string, error) {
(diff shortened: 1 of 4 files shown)


colinmarc:
Huh, interesting. Completely out of curiosity, what's the situation where you're cross-compiling?

colinmarc:
Also, is this bug still true as of 1.5? I know that a lot of the cross-compilation stuff changed.

jpellerin:
FWIW I just ran into this too. 1.4, though -- IDK whether it is fixed in 1.5.

ebartels:
@colinmarc 

```
user: Current not implemented on linux/amd64
```
is there error I see for a binary built on darwin/amd64 using cross compilation?  Can someone else reproduce this?

thomaso-mirodin:
@ebartels I can reproduce this on both 1.5.1 and HEAD

colinmarc:
Since it returns an error in 1.5, how about trying `user.Current()` first and using `GetEnv` if that returns an error? Throw in a comment describing why, and I'll merge it.

ebartels:
The actual Hadoop semantics though are to use` HADOOP_USER_NAME` through the environment unless one is using Kerberos so should this also follow those same semantics rather than enforcing `user.Current()`?

https://hadoop.apache.org/docs/stable/api/src-html/org/apache/hadoop/security/UserGroupInformation.AuthenticationMethod.html
```
173      //If we don't have a kerberos user and security is disabled, check
174      //if user is specified in the environment or properties
175      if (!isSecurityEnabled() && (user == null)) {
176        String envUser = System.getenv(HADOOP_USER_NAME);
177        if (envUser == null) {
178          envUser = System.getProperty(HADOOP_USER_NAME);
179        }
180        user = envUser == null ? null : new User(envUser);
181      }
182      // use the OS user
183      if (user == null) {
184        user = getCanonicalUser(OS_PRINCIPAL_CLASS);
185        if (LOG.isDebugEnabled()) {
186          LOG.debug("using local user:"+user);
187        }
188      }
```
So how about if I change it to default to `GetEnv` and if that isn't passed then it falls back to `user.Current()`?

colinmarc:
@ebartels that makes sense to me, if it's `HADOOP_USER_NAME` instead of `USER`.

SuperQ:
:+1: It would be nice to have a release of this :-)

colinmarc:
Thanks for the patch! Sorry about the delay (and the pickiness) =)

Issue #19, opened 2015-11-22
ReadAt does not have io.ReaderAt behavior


ashishgandhi:
According to the `io.ReaderAt` documentation (https://golang.org/pkg/io/#ReaderAt). For the signature

```
ReadAt(p []byte, off int64) (n int, err error)
```

“ReadAt reads len(p) bytes into p starting at offset off in the underlying input source. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered.

When ReadAt returns n < len(p), it returns a non-nil error explaining why more bytes were not returned. In this respect, ReadAt is stricter than Read.”

If you simply try playing with a file you'd quickly observe that this is not followed. Taking a real world example I have a file of size 280689401. I pass in `p` of length 227743. But only 214 bytes are read. And the `err` is `nil`.

PR #20, opened 2015-11-22
Implement io.ReaderAt correctly

ashishgandhi:
Solves issue https://github.com/colinmarc/hdfs/issues/19.

### DIFF
--- a/cmd/hdfs/cat.go
+++ b/cmd/hdfs/cat.go
@@ -136,7 +136,7 @@ func tailLines(file *hdfs.FileReader, numLines int64) {
 			b, err = section.ReadByte()
 		}
 
-		if err != nil && err != io.EOF {
+		if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
 			fatal(err)
 		}
 
--- a/cmd/hdfs/cat.go
+++ b/cmd/hdfs/cat.go
@@ -4,12 +4,13 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
-	"github.com/colinmarc/hdfs"
 	"io"
 	"os"
+
+	"github.com/colinmarc/hdfs"
 )
 
-var tailSearchSize int64 = 16384
+const tailSearchSize int64 = 16384
 
 func cat(paths []string) {
 	expanded, client, err := getClientAndExpandedPaths(paths)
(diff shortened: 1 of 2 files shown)


ashishgandhi:
I'd love to look into why the tests are failing. So far I have been able to successfully use this patch.

colinmarc:
Thanks for the patch! I'll have to think about the ReaderAt thing. Do you mind putting the import rearrangements into a separate PR? +1 on them, but this diff is super noisy.

colinmarc:
I think the tail implementation is actually wrong, too. `SectionReader` uses `ReadAt`, so that's why the behavior broke. `tail` works with long files, but not ones shorter than the period it's trying to tail.

ashishgandhi:
So, if anything, `SectionReader` should work correctly with this `ReadAt` fix? Also, I got rid of the `goimports` changes in this PR.

ashishgandhi:
I see. The tail implementation is wrong. I'll see how I can fix it. I don't use tail for now so I'll maintain this patch in my fork.

ashishgandhi:
Never mind, tail was not incorrect. I ended up implementing my own only to realize that's not where the problem is. The problem is that `io.ReadFull` returns `io.ErrUnexpectedEOF` while `io.ReadAt` does not. Fixed my fix. Travis is happy and so am I. :smile:

cmd/hdfs/cat.go
```diff
@@ -136,7 +136,7 @@ func tailLines(file *hdfs.FileReader, numLines int64) {
 			b, err = section.ReadByte()
 		}
 
-		if err != nil && err != io.EOF {
+		if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
```
colinmarc: Rather than catching this here, can we tell the SectionReader to only read at most the number of bytes in the file? I think ErrUnexpectedEOF could also happen if something goes wrong with the connection.

ashishgandhi: Done.

ashishgandhi:
I think this is ready for review number 2.

colinmarc:
Perfect. Thanks for the thorough contribution!

ashishgandhi:
Thanks for the library. :smile:

PR #25, opened 2016-01-06
Prevent race condition

sediah:
TestWriteRaceCondition will be failed before applying the patch.


colinmarc:
So, if I understand correctly - this issue happens when `Write` returns an `ackError`, and then the calling code calls `Write` again? Shouldn't the `blockWriteStream` just be considered broken at that point?

But wouldn't this code in `Write` just keep returning the `ackError`, preventing us from getting to the race? I don't quite see it yet.

```go
if s.ackError != nil {
	return 0, s.ackError
}
```

colinmarc:
Ahhhhh, I see it now. `s.ackError` hasn't been set at that point, and then we quit the acking thread without closing the channel.

colinmarc:
I went a slightly different direction with the fix - hope you don't mind. Thanks so much for catching this!

sediah:
Thanks! I'll try with this version.

PR #26, opened 2016-01-06
Simple support for HA

sediah:
Checked on CDH 5.3.x.


colinmarc:
Thanks for the work on this! Zookeeper isn't the only way that HA is set up, right? I was doing a bit of research earlier - I wonder if we can support this without connecting to zookeeper, like the java client does.

colinmarc:
Also, did you forget to include the pb file?

sediah:
Zookeeper is only for automatic failover.
Simpler way without ZK is testing for 'org.apache.hadoop.ipc.StandbyException' from the name node with Stat() or other operations.
Actually I also check this and reconnect.

Issue #28, opened 2016-01-23
TCP connection to data nodes are actively leaked when Seek() or ReadAt() is used


vlivan-microsoft:
*Symphom:* "too many open files" error
*Repro Steps:* open large file and do a bunch of ReadAt() at random positions in a loop
*Details:* block reader isn't closed properly, so active TCP connections might grow with the high rate. Go garbage collector doesn't "feel" memory pressure - so it isn't cleaning them up on time leading to quick exhausting of available fds.

vlivan-microsoft:
Fix: https://github.com/colinmarc/hdfs/pull/29

colinmarc:
Fixed in #29.

PR #29, opened 2016-01-23
fix for #28: properly closing BlockReader on Seek()

vlivan-microsoft:


file_reader.go
```diff
@@ -125,9 +125,13 @@ func (f *FileReader) Seek(offset int64, whence int) (int64, error) {
 		return f.offset, fmt.Errorf("Invalid resulting offset: %d", off)
 	}
 
-	f.offset = off
-	f.blockReader = nil
-
+	if f.offset != off {
+		f.offset = off
+		if f.blockReader != nil {
```
colinmarc: I think we want this to be a separate conditional, so we close it even if `f.offset == off`

colinmarc: Oops, nevermind, I think you were right all along.

PR #32, opened 2016-02-13
Fix ls -l format

SuperQ:
`ls -l` prints Month Day incorrectly for 2-digit dates.  Use the _ empty
padding method to format the Month and Day.


colinmarc:
Ah word! Thanks for the fix.

PR #36, opened 2016-04-03
Format ls -l year correctly

dajobe:
Without this fix, ls -l generated Unicode character 2015 for the year 2015 rather than the string "2015"


colinmarc:
Hah, oops. Thanks for the fix.

PR #37, opened 2016-04-04
Update protocol support for Hadoop 2.7.2

dajobe:
Slightly tested; updates the protocol to the latest release.

Taken from hadoop git repository mirror at https://github.com/apache/hadoop for 2.7.2 released 2016-01-25 from branch 'branch-2.7.2' at commit b165c4fe8a74265c792ce23f546c64604acf0e41

The "Fixup" commit 3b6a5b5 I am a little unsure about; I tried to restore what was previously there.


protocol/Makefile
```diff
@@ -0,0 +1,3 @@
+all:
+	protoc --go_out=hadoop_hdfs/ ClientNamenodeProtocol.proto datatransfer.proto hdfs.proto
```
colinmarc: We can actually target the generated files themselves, right?

colinmarc:
Looks like the imports are still slightly confused.

Thanks for this! It seems backwards compatible, so what's the motivation? Any particular features you'd like to see taken advantage of?

dajobe:
The original code worked but my motivation was to update it and to understand how to build / rebuild it.  e.g. there are hardcoded references to the repo in the generated files.

colinmarc:
Ah yeah. I couldn't figure out how to get the proto generation to play nicely. Props if you can sort that out!

dajobe:
@colinmarc so the latest commits automates fixing the proto generation and the tests pass.  It's better than the first version.  I should try it out on a real hadoop too...

colinmarc:
@dajobe I fought with this for a while today - how about something like this? Should be a bit cleaner than munging imports (either manually or with sed). https://github.com/colinmarc/hdfs/pull/39

PR #38, opened 2016-04-06
Add globbing support

crozzy:
This is loosely based on the snakebite glob support.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,12 +38,15 @@ func New(address string) (*Client, error) {
 		return nil, err
 	}
 
-	return NewForUser(address, username)
-}
+	if address == "" {
+		var nnErr error
+		address, nnErr = GetNamenodeFromConfig()
 
-// Return a New Client with auto config
-func AutoConfigClient() (*Client, error) {
-	return GetAutoConfigClient()
+		if nnErr != nil {
+			return nil, nnErr
+		}
+	}
+	return NewForUser(address, username)
 }
 
 // NewForUser returns a connected Client with the user specified, or an error if
--- a/client.go
+++ b/client.go
@@ -41,6 +41,11 @@ func New(address string) (*Client, error) {
 	return NewForUser(address, username)
 }
 
+// Return a New Client with auto config
+func AutoConfigClient() (*Client, error) {
+	return GetAutoConfigClient()
+}
+
 // NewForUser returns a connected Client with the user specified, or an error if
 // it can't connect.
 func NewForUser(address string, user string) (*Client, error) {
--- a/client_auto.go
+++ b/client_auto.go
@@ -1,0 +1,113 @@
+package hdfs
+
+import (
+	"encoding/xml"
+	"io/ioutil"
+	"net/url"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/colinmarc/hdfs/rpc"
+)
+
+type Property struct {
+	Name  string `xml:"name"`
+	Value string `xml:"value"`
+}
+
+type Result struct {
+	Property []Property `xml:"property"`
+}
+
+type NameNode struct {
+	Host string
+	Port int
+}
+
+// Get Hadoop Properties
+func GetHadoopProperties(path string) ([]Property, error) {
+	result := Result{}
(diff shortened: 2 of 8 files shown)


colinmarc:
Hi! is there any reason `path.Match` doesn't work? I use that in the command line client, which [already supports globs](https://github.com/colinmarc/hdfs/blob/master/cmd/hdfs/test/glob.bats) (implementation [here](https://github.com/colinmarc/hdfs/blob/master/cmd/hdfs/paths.go#L82-L146).

crozzy:
Hi! I guess I didn't think about using it just because I'm comparing 2 strings not paths. But path.Match does make more sense in the way it handles globs.

crozzy:
I actually started using this today and as the go `os.FileInfo` doesn't have a path type it made more sense in my use case to return a `[]string` of matching paths.

client.go
```diff
@@ -41,6 +41,11 @@ func New(address string) (*Client, error) {
 	return NewForUser(address, username)
 }
 
+// Return a New Client with auto config
+func AutoConfigClient() (*Client, error) {
```
colinmarc: Rather than a new method, let's just have this be default fallback behavior for `New` if an empty string is passed for the namenode.

client_auto.go
```diff
@@ -0,0 +1,131 @@
...
+	Message string
+}
+
+func (e *nameNodeError) Error() string {
+	return fmt.Sprintf("%s", e.Message)
+}
+
+// Get Hadoop Properties - try to open a conf file, marshal the results
+// into a Result object and return the Properties of that object.
+func GetHadoopProperties(path string) ([]Property, error) {
```
colinmarc: Let's flatten this into a map, and return that. something like:

```
type HadoopConf map[string]string
func LoadHadoopConf(path string) // if path is empty, search the regular search path
func (conf HadoopConf) Namenodes() []string
```

client_auto.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Get Namenode server name(s) from site config
+func GetNamenodesFromSiteConfig(path string) []string {
+	props, err := GetHadoopProperties(path)
+	if err != nil {
+		return []string{}
+	}
+	var nns []string
+	for _, prop := range props {
+		if strings.Contains(prop.Name, "fs.defaultFS") || strings.Contains(prop.Name, "fs.defaultFS") {
```
colinmarc: both sides of this check are the same (I'm guessing you meant to check for the deprecated `dfs.namenode.rpc-address`). I think that means the method above this is superfluous.

client_auto.go
```diff
@@ -0,0 +1,131 @@
...
+// AutoConfigClient to create a client by trying to read the hadoop config
+// and returning the first if no namenodes are found look for HADOOP_NAMENODE env var
+func GetAutoConfigClient() (*Client, error) {
+	hadoopHome := os.Getenv("HADOOP_HOME")
+	hadoopConfDir := os.Getenv("HADOOP_CONF_DIR")
+	var tryPaths []string
+	var confPaths []string
+	if hadoopHome != "" {
+		hdfsPath := path.Join(hadoopHome, "conf", "hdfs-site.xml")
+		tryPaths = append(tryPaths, hdfsPath)
```
colinmarc: why the difference in `tryPaths` and `corePaths`? I think this could be a bunch cleaner.

colinmarc:
Regarding globs: I don't really see the benefit to including globs in the library, since the command line client already supports them and they're really a bash thing. I'd cite that there's no equivalent method in the go `os` package. A user could easily use path.Match to implement something like you have here if they really want to.

I'm interested in this config loading stuff, though!

client_auto.go
```diff
@@ -0,0 +1,131 @@
...
+type Result struct {
+	Property []Property `xml:"property"`
+}
+
+type NameNode struct {
+	Host string
+	Port int
+}
+
+type nameNodeError struct {
```
colinmarc: There's no need for this to be a struct. it can just be `var ErrUnresolvedNamenode = errors.New("Couldn't find a namenode address in any config.")`

PR #41, opened 2016-04-11
Crozzy auto config

crozzy:
Hey @colinmarc , I think this is a little cleaner. Let me know if you have some other suggestions.


crozzy:
also, feel free to reject #38

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/xml"
+	"errors"
+	"io/ioutil"
+	"net/url"
+	"os"
+	"path"
+	"strings"
+)
+
+type Property struct {
```
colinmarc: This and `Result` don't need to be capitalized, since they aren't exported. Also I would name `Result` something like `propertyList`.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+
+type Result struct {
+	Property []Property `xml:"property"`
+}
+
+type HadoopConf map[string]string
+
+var ErrUnresolvedNamenode = errors.New("Couldn't find a namenode address in any config.")
+
+// Get Hadoop Properties - try to open a conf file, marshal the results
```
colinmarc: Comments for methods should be in the standard go form "LoadHadoopConfig loads the hadoop config from the given path..." etc.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+func LoadHadoopConfig(path string) (HadoopConf, error) {
+	result := Result{}
+	f, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	xmlErr := xml.Unmarshal(f, &result)
+	if xmlErr != nil {
+		return nil, xmlErr
```
colinmarc: you should wrap this error so that the caller gets more information. something like `fmt.Errorf("reading %s: %s", path, xmlErr)` would do it.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	hadoopConf := make(HadoopConf)
```
colinmarc: super minor, but you can save some allocations if you pass the size argument in here, since we know it in advance: `make(HadoopConf, len(result))`

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, prop := range result.Property {
+		hadoopConf[prop.Name] = prop.Value
+	}
+	return hadoopConf, nil
+}
+
+func (conf HadoopConf) Namenodes() []string {
```
colinmarc: This needs a comment, since it's exported.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+type Property struct {
+	Name  string `xml:"name"`
+	Value string `xml:"value"`
+}
+
+type Result struct {
+	Property []Property `xml:"property"`
+}
+
+type HadoopConf map[string]string
```
colinmarc: This needs a comment, since it's exported.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+	if hadoopConfDir != "" {
+		confHdfsPath := path.Join(hadoopConfDir, "hdfs-site.xml")
+		confCorePath := path.Join(hadoopConfDir, "core-site.xml")
+		tryPaths = append(tryPaths, []string{confHdfsPath, confCorePath}...)
+	}
+	var nameNodes []string
+	for _, tryPath := range tryPaths {
+		hadoopConf, err := LoadHadoopConfig(tryPath)
+		if err == nil {
+			nameNodes = append(nameNodes, hadoopConf.Namenodes()...)
```
colinmarc: I wonder if we should just return once we find the config, rather than concatenating the results from all the config files. What do you think?

crozzy: I was wondering the same, but figured all the nodes might be useful for adding some kind of namenode-failover support. I'm happy with either, we can be pretty sure we will loop a max of 4 times so performance increases will be small. Up to you

client_auto.go
```diff
@@ -0,0 +1,95 @@
+package hdfs
```
colinmarc: This should probably just be `conf.go`.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+	Property []Property `xml:"property"`
+}
+
+type HadoopConf map[string]string
+
+var ErrUnresolvedNamenode = errors.New("Couldn't find a namenode address in any config.")
+
+// Get Hadoop Properties - try to open a conf file, marshal the results
+// into a Result object and return the Properties of that object.
+func LoadHadoopConfig(path string) (HadoopConf, error) {
```
colinmarc: Let's stick with `conf` instead of `config` in the name, for consistency.

The `HADOOP_HOME`/`HADOOP_CONF_DIR` stuff below is potentially useful for other properties, too. How about we have that be the default behavior for this method, if path is an empty string? Then we don't need `GetNamenodeFromConfig` at all.

client.go
```diff
@@ -38,6 +38,14 @@ func New(address string) (*Client, error) {
 		return nil, err
 	}
 
+	if address == "" {
```
colinmarc: This looks good but we should update the documentation for the method.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+func (conf HadoopConf) Namenodes() []string {
+	var nns []string
+	for key, value := range conf {
+		if strings.Contains(key, "fs.defaultFS") {
+			nnUrl, _ := url.Parse(value)
+			nns = append(nns, nnUrl.Host)
+		}
+		if strings.HasPrefix(key, "dfs.namenode.rpc-address") {
+			nns = append(nns, value)
+		}
```
colinmarc: This method should dedupe, I think. You can do that by writing to a `map[string]bool` first, and then pulling out the keys into a slice to return.

colinmarc:
Ok - left a bazillion comments, sorry =) Thanks again for your work on this! I'll close #38.

client_auto.go
```diff
@@ -0,0 +1,95 @@
...
+	Value string `xml:"value"`
+}
+
+type Result struct {
+	Property []Property `xml:"property"`
+}
+
+type HadoopConf map[string]string
+
+var ErrUnresolvedNamenode = errors.New("Couldn't find a namenode address in any config.")
```
colinmarc: Go errors should be shorthand, uncapitalized and unpunctuated, like `"no namenode address in config"`.

client_auto.go
colinmarc: Yeah, not worried about the perf, just the semantics. I think we should keep this behavior (load everything) but put it in the method above, like I was saying. That way you end up with one big `HadoopConf` that has all the properties from all the files. We just need to pick a resolution order (`HADOOP_CONF_DIR` before `HADOOP_HOME`, I guess?) and stick with it.

client_auto.go
crozzy: Yeah, it'd be good to put the behavior in `LoadHadoopConfig` although I think `GetNamenodeFromConfig` is useful for keeping the logic that checks `HADOOP_NAMENODE` if no namenode is found in the hadoop config.

colinmarc: hm is `HADOOP_NAMENODE` a real hadoop property? If not, we can just skip that.

crozzy:
No problem! Let know if there are some more changes for this PR.

client.go
```diff
@@ -31,16 +31,37 @@ func Username() (string, error) {
...
 
+// getNameNodeFromConf return a datanode from the system hadoop configuration
+func getNameNodeFromConf() (string, error) {
+	hadoopConf := LoadHadoopConf("")
+
+	namenodes, nnErr := hadoopConf.Namenodes()
+	if nnErr != nil {
+		return "", nnErr
+	}
+	return namenodes[len(namenodes)-1], nil
```
crozzy: Maybe this should be `namenodes[0]`

conf.go
```diff
@@ -0,0 +1,98 @@
...
+		hadoopHome := os.Getenv("HADOOP_HOME")
+		if hadoopConfDir != "" {
+			confHdfsPath := path.Join(hadoopConfDir, "hdfs-site.xml")
+			confCorePath := path.Join(hadoopConfDir, "core-site.xml")
+			tryPaths = append(tryPaths, []string{confHdfsPath, confCorePath}...)
+		}
+		if hadoopHome != "" {
+			hdfsPath := path.Join(hadoopHome, "conf", "hdfs-site.xml")
+			corePath := path.Join(hadoopHome, "conf", "core-site.xml")
+			tryPaths = append(tryPaths, []string{hdfsPath, corePath}...)
```
colinmarc: this can just be `append(tryPaths, confHdfsPath, confCorePath)`

conf.go
```diff
@@ -0,0 +1,98 @@
...
+	} else {
+		hadoopConfDir := os.Getenv("HADOOP_CONF_DIR")
+		hadoopHome := os.Getenv("HADOOP_HOME")
+		if hadoopConfDir != "" {
+			confHdfsPath := path.Join(hadoopConfDir, "hdfs-site.xml")
+			confCorePath := path.Join(hadoopConfDir, "core-site.xml")
+			tryPaths = append(tryPaths, []string{confHdfsPath, confCorePath}...)
+		}
+		if hadoopHome != "" {
+			hdfsPath := path.Join(hadoopHome, "conf", "hdfs-site.xml")
```
colinmarc: I think you want `filepath.Join` instead of `path.Join`, since it's on the local filesystem.

client.go
colinmarc: Yeah I guess that's slightly better (but if we have multiple I'm not really sure what the behavior should be, outside of HA support).

conf.go
crozzy: Is the only difference that filepath supports windows filesystems?

colinmarc: Yeah, `path/filepath` is OS-agnostic but `path` assumes `/`.

colinmarc:
Ok, looks awesome. A couple more minor changes and then I'll be happy to merge!

crozzy:
Sweet! Thanks for the tips.

colinmarc:
Yay! Thank you so much for the PR.

PR #45, opened 2016-07-01
Fixed incorrect Append implementation.

tyler-sommer:
The previous implementation considered existing blocks immutable, which is not
correct. It is possible to append to an existing block by unfinalizing it prior to writing.

This commit attempts to unfinalize the last existing block during an append
operation so that it may be written to again. This is done by submitting the
correct stage `hdfs.OpWriteBlockProto_PIPELINE_SETUP_APPEND` during the
initial OpWriteBlock operation. Additionally, `blockWriteStream`s are now
always initialized with an offset, which will be > 0 in case of append
operation. Finally, the GenerationStamp field is incremented at the beginning
of the append operation, which is an incrementing integer that records the
version of a particular block.

### DIFF
--- a/file_writer.go
+++ b/file_writer.go
@@ -90,8 +90,8 @@ func (c *Client) Append(name string) (*FileWriter, error) {
 	}
 
 	appendReq := &hdfs.AppendRequestProto{
-		Src:          proto.String(name),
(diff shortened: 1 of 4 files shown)


colinmarc:
Sorry for missing this before! Looks great.

Issue #48, opened 2016-08-17
CLI user


binary132:
Setting the HADOOP_USER_NAME variable isn't documented in the README.  I'll open a PR momentarily.

PR #49, opened 2016-08-17
Document HADOOP_USER_NAME in README.md

binary132:
Address Issue #48.


colinmarc:
I went with slightly different wording - fixed in b482024. Thanks for the suggestion!

PR #53, opened 2016-09-23
Don't increment the generation timestamp when appending

colinmarc:
This adds a test covering an error the namenode returns when we try to append twice to the existing block, and a fix for it (which I confess to not really understanding).

@tyler-sommer, this seems to directly contradict the comment you left, so what do you think? Does this seem right?

### DIFF
--- a/file_writer_test.go
+++ b/file_writer_test.go
@@ -304,3 +304,46 @@ func TestFileAppendLastBlockFull(t *testing.T) {
 
 	assert.Equal(t, "\nfoo", string(buf))
 }
+
+func TestFileAppendTwice(t *testing.T) {
+	client := getClient(t)
+
+	baleet(t, "/_test/append/4.txt")
+	mkdirp(t, "/_test/append")
+	writer, err := client.Create("/_test/append/4.txt")
+	require.NoError(t, err)
+
+	n, err := writer.Write([]byte("foo"))
+	require.NoError(t, err)
+	assert.Equal(t, 3, n)
+
+	err = writer.Close()
+	require.NoError(t, err)
+
+	writer, err = client.Append("/_test/append/4.txt")
+	require.NoError(t, err)
+
+	n, err = writer.Write([]byte("bar"))
+	require.NoError(t, err)
+	assert.Equal(t, 3, n)
+
+	err = writer.Close()
+	require.NoError(t, err)
+
+	writer, err = client.Append("/_test/append/4.txt")
+	require.NoError(t, err)
+
+	n, err = writer.Write([]byte("baz"))
+	require.NoError(t, err)
+	assert.Equal(t, 3, n)
+
+	err = writer.Close()
+	require.NoError(t, err)
+
+	reader, err := client.Open("/_test/append/4.txt")
+	require.NoError(t, err)
+
+	bytes, err := ioutil.ReadAll(reader)
+	require.NoError(t, err)
+	assert.Equal(t, "foobarbaz", string(bytes))
+}
--- a/rpc/block_writer.go
+++ b/rpc/block_writer.go
@@ -160,10 +160,8 @@ func (bw *BlockWriter) currentStage() hdfs.OpWriteBlockProto_BlockConstructionSt
 }
 
 func (bw *BlockWriter) generationTimestamp() int64 {
-	// TODO: This needs to be incremented only when appending to
-	// an existing block.
 	if bw.append {
-		return int64(bw.block.B.GetGenerationStamp() + 1)
+		return int64(bw.block.B.GetGenerationStamp())
 	}
 	return 0
 }


tyler-sommer:
@colinmarc, It appears the client is supposed to let the NameNode increment the generation stamp. Long story short, the `updateBlockForPipeline` rpc method takes an existing block and will increase the generation stamp when necessary.

## Full explanation

Looking into it, it seems that the generation stamp should indeed be incremented during append operations.

Check out this [article by cloudera about HDFS recovery processes](http://blog.cloudera.com/blog/2015/02/understanding-hdfs-recovery-processes-part-1/). ([Part two is also interesting](http://blog.cloudera.com/blog/2015/03/understanding-hdfs-recovery-processes-part-2/))

In the article, it states:

> **Generation Stamp**
> A GS is a monotonically increasing 8-byte number for each block that is maintained persistently by the NameNode. The GS for a block and replica is introduced for the following purposes:

> * Detecting stale replica(s) of a block: that is, when the replica GS is older than the block GS, which might happen when, for example, an append operation is somehow skipped at the replica.
> * Detecting outdated replica(s) on DataNodes which have been dead for long time and rejoin the cluster.

> A new GS is needed when any of the following occur:

> * A new file is created
> * A client opens an existing file for append or truncate
> * A client encounters an error while writing data to the DataNode(s) and requests a new GS
> * The NameNode initiates lease recovery for a file

[You can see here in the BlockReceiver HDFS source code](https://github.com/apache/hadoop/blob/ac0538aac347bfd97cc0dee1db49db503c15f1d9/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/datanode/BlockReceiver.java#L201) that Append operations sort of expect a new GS being passed.


### Understanding the Java client implementation

I've retraced the steps that the Java client takes when writing to blocks and here they are:

1. The whole append operation [starts with creating a new DFSOutputStream](https://github.com/apache/hadoop/blob/ac0538aac347bfd97cc0dee1db49db503c15f1d9/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DFSOutputStream.java#L1258). You can see that when this is called, a new GS is passed along. The method `createBlockOutputStream` is called before writing to each block, in any mode (append, create, or recovery).

    1. For example, [here `createBlockOutputStream` is called](https://github.com/apache/hadoop/blob/ac0538aac347bfd97cc0dee1db49db503c15f1d9/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DFSOutputStream.java#L1237) when starting the write to a brand new block, note the `0L` as the newGS.

    2. More interestingly for our case, [here is where `createBlockOutputStream` is called](https://github.com/apache/hadoop/blob/ac0538aac347bfd97cc0dee1db49db503c15f1d9/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DFSOutputStream.java#L1159) during Append or Recovery operations.

2. When the client is appending, [a new generation stamp is retrieved from the namenode](https://github.com/apache/hadoop/blob/ac0538aac347bfd97cc0dee1db49db503c15f1d9/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DFSOutputStream.java#L1144) using the `updateBlockForPipeline` method which is, very handily, exposed by the protobuffer API.


### Continuing further, mostly as a matter of curiosity

These steps just show how the namenode actually goes about increasing the generation stamp. Not necessarily pertinent, but maybe interesting.

1. [The namenode "bumps" the generation stamp.](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/NameNodeRpcServer.java#L913)

2. [Which is implemented by the FSNameSystem here](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSNamesystem.java#L4883)

3. [Which in turn calls this method](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSNamesystem.java#L4744)

4. [Which calls this method on `BlockManager`](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockManager.java#L4591)

5. [Which delegates to the `BlockIdManager`](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/BlockIdManager.java#L167)

6. [Which calls `nextValue` on `GenerationStamp`](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/common/GenerationStamp.java)

7. [Which is actually implemented on `SequentialNumber`](https://github.com/apache/hadoop/blob/ea0c2b8b051a2d14927e8f314245442f30748dc8/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/SequentialNumber.java#L49).

### Conclusion

So it seems we should just make an additional call to the `updateBlockForPipeline` RPC endpoint prior to submitting the Append operation. The updated block returned will contain the correct and expected generationStamp as defined by the NameNode.

tyler-sommer:
@colinmarc, I was wrong on all counts. The server handles incrementing the generation stamp. See my edits to my previous comment.

colinmarc:
Awesome, thank you for the detailed writeup! Going to merge this.

PR #55, opened 2016-10-25
Clobber in rename: provide overwrite option in Rename

junjieqian:
Right now, rename api cannot overwrite the existing file, according to https://github.com/colinmarc/hdfs/blob/master/rename.go#L14.

This patch allows users to decide whether to overwrite files if the files already exist.

### DIFF
--- a/cmd/hdfs/mv.go
+++ b/cmd/hdfs/mv.go
@@ -57,14 +57,18 @@ func moveInto(client *hdfs.Client, sources []string, dest string, force bool) {
 }
 
 func moveTo(client *hdfs.Client, source, dest string, force bool) {
+	resp, err := client.Stat(dest)
 	if force {
-		err := client.Remove(dest)
-		if err != nil && !os.IsNotExist(err) {
-			fatal(err)
+		if err == nil && resp.IsDir() {
+			if err = client.Remove(dest); err != nil {
+				fatal(err)
+			}
 		}
+	} else if err == nil {
+		fatal(&os.PathError{"rename", dest, os.ErrExist})
 	}
 
-	err := client.Rename(source, dest)
+	err = client.Rename(source, dest)
 	if err != nil {
 		fatal(err)
 	}
--- a/rename.go
+++ b/rename.go
@@ -11,9 +11,7 @@ import (
 // Rename renames (moves) a file.
 func (c *Client) Rename(oldpath, newpath string) error {
 	_, err := c.getFileInfo(newpath)
-	if err == nil {
-		return &os.PathError{"rename", newpath, os.ErrExist}
-	} else if !os.IsNotExist(err) {
+	if err != nil && !os.IsNotExist(err) {
 		return &os.PathError{"rename", newpath, err}
 	}
 
--- a/rename_test.go
+++ b/rename_test.go
@@ -42,7 +42,7 @@ func TestRenameDestExists(t *testing.T) {
 	touch(t, "/_test/tomovedest2")
 
 	err := client.Rename("/_test/tomove2", "/_test/tomovedest2")
-	assertPathError(t, err, "rename", "/_test/tomovedest2", os.ErrExist)
+	require.NoError(t, err)
 }
 
 func TestRenameWithoutPermissionForSrc(t *testing.T) {


colinmarc:
Hm, I'd rather keep the api the same as the [os equivalent](https://golang.org/pkg/os/#Rename), and preserve backwards compatibility. If the behavior is different from the os version, though, I think that's a bug, and we can just change the default.

junjieqian:
Thanks for the reply!
In that case, we may still need a small fix to rename.go#L14, as it is not same as rename in golang os package, that now it will report os.ErrExist if the dest file exists. Some of tests may need to be modified as well, because the rename returns success even if dest file exists.
Please correct me if I am wrong.

colinmarc:
Yeah, I'm in favor of changing the behavior and tests in this case. We're currently violating the posix spec for `mv`, as well.

cmd/hdfs/mv.go
```diff
@@ -62,6 +63,11 @@ func moveTo(client *hdfs.Client, source, dest string, force bool) {
 		if err != nil && !os.IsNotExist(err) {
```
colinmarc: We no longer need this `Remove`.

rename.go
```diff
@@ -12,7 +12,7 @@ import (
 func (c *Client) Rename(oldpath, newpath string) error {
 	_, err := c.getFileInfo(newpath)
 	if err == nil {
-		return &os.PathError{"rename", newpath, os.ErrExist}
+		// Do nothing, the existing newpath will be overwritten
```
colinmarc: You can just remove this `getFileInfo` altogether.

cmd/hdfs/mv.go
```diff
@@ -5,6 +5,7 @@ import (
 	"path"
 
 	"github.com/colinmarc/hdfs"
+	"github.com/colinmarc/hdfs/protocol/hadoop_hdfs"
 )
 
 func mv(paths []string, force, treatDestAsFile bool) {
```
colinmarc: I think the functionality of `mv` should change, too. On my machine, `mv` overwrites by default.

rename.go
junjieqian: I kept this `getFileInfo`, as we need this to verify if we have permission to write to newpath.

cmd/hdfs/mv.go
junjieqian: I updated changes accordingly.

junjieqian:
Thank you very much for your comments! I have addressed them accordingly. I replied some comments that why I think they are needed, please let me know if you think another way is better.

I will combine two commits into one, once it is approved.

cmd/hdfs/test/mv.bats
```diff
@@ -89,7 +89,7 @@ OUT
   run $HDFS mv -T /_test_cmd/mv/a /_test_cmd/mv/dir
   assert_failure
   assert_output <<OUT
-rename /_test_cmd/mv/dir: file already exists
+rename /_test_cmd/mv/dir file already exists
```
colinmarc: Why did these change?

colinmarc:
Thanks, I think we're nearly there! Just a few nitpicks.

cmd/hdfs/mv.go
```diff
@@ -57,14 +57,21 @@ func moveInto(client *hdfs.Client, sources []string, dest string, force bool) {
 }
 
 func moveTo(client *hdfs.Client, source, dest string, force bool) {
+	resp, err := client.Stat(dest)
 	if force {
-		err := client.Remove(dest)
-		if err != nil && !os.IsNotExist(err) {
-			fatal(err)
+		// If dest exists and is directory, remove dest directory
```
colinmarc: This comment doesn't add much; the code is pretty self-explanatory.

cmd/hdfs/mv.go
```diff
@@ -57,14 +57,21 @@ func moveInto(client *hdfs.Client, sources []string, dest string, force bool) {
...
+		if err == nil && resp.IsDir() {
+			if err = client.Remove(dest); err != nil {
+				fatal(err)
+			}
+		}
+	} else {
+		if err == nil {
+			fatal("rename", dest, os.ErrExist)
```
colinmarc: This should be a `os.PathError`, I think

junjieqian: Yes, you are right! I changed this line to `os.PathError`, and so I undid the changes in `cmd/hdfs/test/mv.bat`.

junjieqian:
Hi Colin, I removed the comments as instructed. The changes in `cmd/hdfs/test/mv.bats` are caused by not using `os.PathError` in `cmd/hdfs/mv.go`, which are removed now.
Again thank you very much for your comments!

cmd/hdfs/mv.go
```diff
@@ -57,14 +57,20 @@ func moveInto(client *hdfs.Client, sources []string, dest string, force bool) {
...
 	if force {
-		err := client.Remove(dest)
-		if err != nil && !os.IsNotExist(err) {
-			fatal(err)
+		if err == nil && resp.IsDir() {
+			if err = client.Remove(dest); err != nil {
+				fatal(err)
+			}
+		}
+	} else {
```
colinmarc: this can be `else if err != nil`

colinmarc:
One more comment =) Then this is ready to merge. Thanks so much for the contribution!

junjieqian:
Thank you for your comments, :-)! Please do let me know if you think there is anything else can be improved.

colinmarc:
Thanks again!

PR #59, opened 2016-11-16
add option like fs -count

hikoz:
I know it's not POSIX option. but I need it.

### DIFF
--- a/cmd/hdfs/du.go
+++ b/cmd/hdfs/du.go
@@ -10,7 +10,7 @@ import (
 	"github.com/colinmarc/hdfs"
 )
 
-func du(args []string, summarize, humanReadable bool) {
+func du(args []string, summarize, humanReadable, withCount bool) {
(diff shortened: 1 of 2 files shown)


colinmarc:
I don't feel particularly good about adding this to `du` - as you said, it's not in the posix spec. 

Programmatically, you can get the total count of a directory in HDFS with the [`GetContentSummary`  method](https://github.com/colinmarc/hdfs/blob/cefa4bc7e7d2f3d849179d6ebe7b0b4d52b3507c/content_summary.go#L59-L63). Could you just write a small go binary for your own infra that exposes that, for your purposes? 

Alternatively, something like `hdfs summary` maybe would work. By default it could display a human readable format, and then with `-s` just the size, `-c` just the count, etc.

hikoz:
Thank you for your reply.
I'll write new command like `summary` as you said.

PR #63, opened 2016-12-06
new client consturctor with rpc.NamenodeConnection 

### LINKED ISSUE #62: allow to set rpc.NamenodeConnection for the hdfs client
in my case I use a Socks proxy to access hdfs. I can already use the `rpc.WrapNamenodeConnection` to create an `*rpc.NamenodeConnection` but I cannot use it for the client.

This would work:
```
func NewForNamenode(namenode *rpc.NamenodeConnection) *Client {
	return &Client{namenode: namenode}
}
```

If if fits the project should I submit a PR ?

advincze:
fixes #62

### DIFF
--- a/client.go
+++ b/client.go
@@ -73,6 +73,12 @@ func NewForUser(address string, user string) (*Client, error) {
 	return &Client{namenode: namenode}, nil
 }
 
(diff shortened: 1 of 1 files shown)


colinmarc:
Thanks for the contribution!

PR #67, opened 2017-02-13
Allow the CLI to load the NameNodes from the Hadoop configuration

sakserv:


colinmarc:
Thanks for the patch!

PR #69, opened 2017-03-10
stat fs info and df command

junjieqian:
This feature reports HDFS usage and provides support for `df` command (result as `$HDFS fs -df (-h)`). Is there a good way to test `df` in `bat` file? Now, the commit just check if it is success.

### DIFF
--- a/cmd/hdfs/df.go
+++ b/cmd/hdfs/df.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/colinmarc/hdfs"
+)
+
+func df(humanReadable bool) {
+	client, err := getClient("")
+	if err != nil {
+		fatal(err)
+	}
+
+	var fs hdfs.FsInfo
+
+	fs, err = client.StatFs()
+	if err != nil {
+		fatal(err)
+	}
+
+	tw := dfTabWriter()
+	fmt.Fprintf(tw, "Filesystem \tSize \tUsed \tAvailable \t Use\n")
+	if humanReadable {
+		fmt.Fprintf(tw, "%v \t%v \t%v \t%v \t%d%%\n",
+			os.Getenv("HADOOP_NAMENODE"),
+			formatBytes(fs.Capacity()),
+			formatBytes(fs.Used()),
+			formatBytes(fs.Remaining()),
+			100 * fs.Used() / fs.Capacity())
+	} else {
+		fmt.Fprintf(tw, "%v \t%v \t %v \t %v \t%d%%\n",
+			os.Getenv("HADOOP_NAMENODE"),
+			fs.Capacity(),
+			fs.Used(),
+			fs.Remaining(),
+			100 * fs.Used() / fs.Capacity())
+	}
+	tw.Flush()
+}
+
+func dfTabWriter() *tabwriter.Writer {
+	return tabwriter.NewWriter(os.Stdout, 3, 8, 0, ' ', tabwriter.AlignRight)
+}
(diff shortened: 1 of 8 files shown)


cmd/hdfs/df.go
```diff
@@ -0,0 +1,46 @@
...
+
+	var fs hdfs.FsInfo
+
+	fs, err = client.StatFs()
+	if err != nil {
+		fatal(err)
+	}
+
+	tw := dfTabWriter()
+	fmt.Fprintf(tw, "Filesystem \tSize \tUsed \tAvailable \t Use\n")
```
colinmarc: This is `Use%` on the version of `df` I have.

cmd/hdfs/df.go
```diff
@@ -0,0 +1,46 @@
...
+			os.Getenv("HADOOP_NAMENODE"),
+			fs.Capacity(),
+			fs.Used(),
+			fs.Remaining(),
+			100 * fs.Used() / fs.Capacity())
+	}
+	tw.Flush()
+}
+
+func dfTabWriter() *tabwriter.Writer {
```
colinmarc: I don't see a reason to have this split out as a separate function - just put it inline.

stat_fs.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"errors"
+
+	hdfs "github.com/colinmarc/hdfs/protocol/hadoop_hdfs"
+)
+
+var StatFsError = errors.New("Failed to get HDFS usage")
+
+// FsInfo provides information about HDFS
+type FsInfo struct {
```
colinmarc: There's no interface to implement, so I'd rather just make these exported and drop the getter methods.

stat_fs.go
```diff
@@ -0,0 +1,66 @@
...
+	corruptBlocks         uint64
+	missingBlocks         uint64
+	missingReplOneBlocks  uint64
+	blocksInFuture        uint64
+	pendingDeletionBlocks uint64
+}
+
+func (c *Client) StatFs() (FsInfo, error) {
+	fs, err := c.getFsInfo()
+	if err != nil {
```
colinmarc: Follow the pattern used in the other methods:

```
if err != nil {
	if nnErr, ok := err.(*rpc.NamenodeError); ok {
		err = interpretException(nnErr.Exception, err)
	}

	return nil, err
}
```

Also, this can just be one method.

stat_fs_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStatFs(t *testing.T) {
+	t.Skip("Stat FS does not work on Travis CI")
```
colinmarc: We can make this work by

a) making all the other tests clean up after themselves (most should already)
b) assuming a fixed size, rather than shelling out

colinmarc: But for now, it's probably fine to just remove the shelling out and make sure the command succeeds at all.

junjieqian:
Hi @colinmarc , thank you for your comments! I fixed them in a new commit. For test, it actually works in stat_fs_test.go (compared with the native `hadoop -fs df` results), but I forgot to remove `skip` in the first commit. The one does not work is `df`, which I am not sure how to best test on Travis.

stat_fs.go
junjieqian: fixed. Thanks

colinmarc:
Ah, I see. Well, that still adds a dependency to hadoop for the golang tests, which we didn't have before - can you change it to just a test for success? Maybe just check that all the values are nonzero.

colinmarc:
Actually, I got it - merged in d9614569203878ff04bb4420b929923949e855fc. Thanks again!

PR #73, opened 2017-03-28
`truncate` support

junjieqian:
This PR is to provide `truncate` feature. About the test:
1. Cloudera's package uses Hadoop-2.6, which does not support `truncate` yet (till Hadoop-2.7, `truncate` is supported).
2. `truncate` takes some time to be effective, and now a time delay (3 seconds, longer maybe safer) is added to the test. Not sure if we can find another better way for this, as the users may need to be aware this delay as well when `truncate` is used.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -189,7 +189,7 @@ func TestCopyToRemote(t *testing.T) {
 
 	baleet(t, "/_test/copytoremote.txt")
 	err := client.CopyToRemote("testdata/foo.txt", "/_test/copytoremote.txt")
-	require.NoError(t, err)
+	ignoreErrReplicating(t, err)
 
 	bytes, err := client.ReadFile("/_test/copytoremote.txt")
 	require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


colinmarc:
Hey @junjieqian,

This is great work, but I'm a bit worried about adding library support for a feature that's barely supported in the actual code - especially if that means we can't test it appropriately. If it's ok, I'd like to wait to merge this PCR until 2.7 is better-supported in most or all the major distros. CDH6 seems like it should bump up to 2.7, at least.

junjieqian:
Hi @colinmarc , sure, let us hold this until 2.7 is supported in all the distros, 👍 .

gotoxu:
Can this PR be merged into the main branch now? We really need this feature.

hykych:
Can this PR be merged into the main branch now? We really need this feature. @colinmarc

colinmarc:
Thanks @junjieqian! Merged in 2f11406.

PR #74, opened 2017-04-05
Change Block Checksum Algorithm to CRC32C

gkalele:
This is to update the block checksum algorithm to the one used by recent Hadoop distributions.

**http://stackoverflow.com/questions/26429360/crc32-vs-crc32c**

### DIFF
--- a/rpc/block_writer.go
+++ b/rpc/block_writer.go
@@ -211,7 +211,7 @@ func (bw *BlockWriter) writeBlockWriteRequest(w io.Writer) error {
 		MaxBytesRcvd:          proto.Uint64(uint64(bw.offset)),
 		LatestGenerationStamp: proto.Uint64(uint64(bw.generationTimestamp())),
 		RequestedChecksum: &hdfs.ChecksumProto{
-			Type:             hdfs.ChecksumTypeProto_CHECKSUM_CRC32.Enum(),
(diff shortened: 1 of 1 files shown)


colinmarc:
Hm - I think the test failures are because we need to change the actual checksum calculation, too.

gkalele:
It looks like the tests run against older Hadoop code perhaps ?

This checksum algorithm needs to be configurable, I guess, to not break compatibility.

The error happens during a Hadoop distcp transfer - files created with Go binaries using this HDFS library fail due to checksum mismatch with files created by standard Hadoop Java binaries.

See:
http://stackoverflow.com/questions/22905411/getting-check-sum-mismatch-during-data-transfer-between-two-different-version-of

JIRA https://issues.apache.org/jira/browse/HDFS-3054

gkalele:
Even though the comments above talk about different versions of Hadoop, we only have 2.4.0.2 installed - files created by Go _appear_ to be created by an older version because of the checksum algorithm.

colinmarc:
The tests run against 2.6, so I don't think it's that. This sounds like a distcp bug, so I'm a little wary of changing anything. 

If we did change it, it would probably work to change the checksum here: https://github.com/gkalele/hdfs/blob/4860b1c776646df8a7bb253b55533fd0dd3428b7/rpc/block_write_stream.go#L202

gkalele:
Needs work, I will close this PR for now.

PR #77, opened 2017-05-08
support for multiple namenodes with failover

itszootime:
Hello! Here's a basic implementation for multiple namenode support.

- If a request results in a non-`NamenodeError` (i.e. EOF) or `StandbyException`, the client will close the current connection, try to establish a new connection with another host, and retry the request.
- The failing host will be marked as down and a connection attempt won't be made for at least 5 seconds (not configurable at the moment, but could be).
- If all hosts were unreachable or in standby, the client request will result in a `"No available namenodes"` error.

I realise there isn't much in terms of testing for this behaviour - would be happy to hear some suggestions for how this could be done.

### DIFF
--- a/client.go
+++ b/client.go
@@ -16,6 +16,13 @@ type Client struct {
 	defaults *hdfs.FsServerDefaultsProto
 }
 
+// ClientOptions represents the configurable options for a client.
+type ClientOptions struct {
+	Addresses []string
+	Namenode  *rpc.NamenodeConnection
+	User      string
+}
+
 // Username returns the value of HADOOP_USER_NAME in the environment, or
 // the current system user if it is not set.
 func Username() (string, error) {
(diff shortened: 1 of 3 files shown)


colinmarc:
This looks like it should work fine, although testing is a concern. The other issue is that we can't change the signature of `New` and `NewNamenodeConnection`. A few other issues require a `NewOptions` method, so that'd be a good place to add this functionality, but I haven't gotten around to implementing that.

ccl0326:
any progress?

itszootime:
I've recently refactored this to support passing options, just need to do a bit more local testing before pushing it up.

itszootime:
@colinmarc only took 3 months, but I've updated the PR. Let me know what you think. Cheers!

client.go
```diff
@@ -30,53 +37,79 @@ func Username() (string, error) {
...
 	if nnErr != nil {
-		return "", nnErr
+		return nil, nnErr
 	}
-	return namenodes[0], nil
+	return namenodes, nil
 }
 
 // NewForUser returns a connected Client with the user specified, or an error if
 // it can't connect.
```
colinmarc: We should probable deprecate these, but that can be a separate PR.

rpc/namenode.go
```diff
@@ -55,45 +67,96 @@ func (err *NamenodeError) Error() string {
...
 
-// WrapNamenodeConnection wraps an existing net.Conn to a Namenode, and preforms
-// an initial handshake.
-//
-// You probably want to use hdfs.New instead, which provides a higher-level
-// interface.
-func WrapNamenodeConnection(conn net.Conn, user string) (*NamenodeConnection, error) {
+// NewNamenodeConnectionWithOptions creates a new connection to a namenode with
+// the given options and performs an initial handshake.
+func NewNamenodeConnectionWithOptions(options NamenodeConnectionOptions) (*NamenodeConnection, error) {
```
colinmarc: Unfortunately, because I'm dumb and made this exported way back when I started this library, we can't just remove the old function. Can you leave it and make it deprecated?

rpc/namenode.go
```diff
@@ -55,45 +67,96 @@ func (err *NamenodeError) Error() string {
...
+		if err != nil {
+			c.markFailure(err)
+			continue
+		}
+
+		break
+	}
+
+	if c.conn == nil {
+		return fmt.Errorf("No available namenodes")
```
colinmarc: `No` shouldn't be capitalized. This can also just be `errors.New`. Better, though, would be to include the last error, so that this isn't completely opaque in degenerate cases: something like `fmt.Errorf("no available namenodes: %s", lastErr)`

rpc/namenode.go
```diff
@@ -55,45 +67,96 @@ func (err *NamenodeError) Error() string {
 	return s
 }
 
-// NewNamenodeConnection creates a new connection to a Namenode, and preforms an
+type host struct {
```
colinmarc: `namenodeHost`?

rpc/namenode.go
```diff
@@ -109,19 +172,28 @@ func (c *NamenodeConnection) Execute(method string, req proto.Message, resp prot
...
-		if _, ok := err.(*NamenodeError); !ok {
-			c.conn.Close() // TODO don't close on RPC failure
+		if nerr, ok := err.(*NamenodeError); ok {
+			if nerr.Exception != "org.apache.hadoop.ipc.StandbyException" {
+				return err
+			}
 		}
-
-		return err
+		c.markFailure(err)
```
colinmarc: This is a bit confusing, but I think this treats `StandbyException`s the same as any non-`NamenodeError` error.

rpc/namenode.go
colinmarc: (also, for the single-namenode case, that would report the error, which is what we want)

rpc/namenode.go
```diff
@@ -109,19 +172,28 @@ func (c *NamenodeConnection) Execute(method string, req proto.Message, resp prot
...
+	err := c.resolveConnection()
 	if err != nil {
-		c.conn.Close()
 		return err
 	}
 
+	err = c.writeRequest(method, req)
+	if err != nil {
+		c.markFailure(err)
+		goto R
```
colinmarc: Elsewhere in the code I use `for` loops and `continue`/`break` rather than goto. This is probably a bit shorter, but I think harder to read.

colinmarc:
Thanks for the changes! The netcode here looks pretty solid, but I'm a bit nervous about not really having a good test framework for it. Has anyone run this in production?

rpc/namenode.go
itszootime: Yep, the operation will only retry if it's either a non-`NamenodeError` or a `StandbyException`

itszootime:
@colinmarc: I think I've implemented all of your suggestions now.

I understand your concerns about testing this. Locally I've been using a [hadoop-ha Docker setup](https://hub.docker.com/r/bastipaeltz/hadoop-ha-docker/) - maybe this would be an option for automated testing too?

We've been using this code in production for a distributed service that mirrors files from HDFS. We haven't had any issues - it handled an active namenode switchover as expected.

colinmarc:
Sorry for the delay. I poked around with this locally and it does seem to do the thing, so I'm going to merge it. I wish we had better tests for the different connection failure modes, but I don't really see a way to do that.

Thanks so much for the contribution!

PR #78, opened 2017-06-29
Set timeout for reads and write to the data nodes.

pareshverma91:
We are facing an issue where the read process gets stuck waiting on the read to complete. The underlying connection was closed from the server side, but the client was still expecting to read something, and hence blocked. Adding timeout to avoid getting stuck.


pareshverma91:
My go constructs are incorrect. Fixing the errors and resubmitting the PR.

PR #82, opened 2017-08-09
Fix alignment in blockWriteStream.

adamfaulkner:
There was a bug that caused packet size to always equal chunkSize (512 bytes). I discovered this when debugging performance with `perf trace`.

Fixed this by taking `alignment = (chunkSize - (offset % chunkSize)) % chunkSize`, which will be 0 when the offset is aligned.

Added a simple test to verify this.

### DIFF
--- a/rpc/block_write_stream.go
+++ b/rpc/block_write_stream.go
@@ -173,7 +173,7 @@ func (s *blockWriteStream) makePacket() outboundPacket {
 	// gets unhappy unless we first align to a chunk boundary with a small packet.
 	// Otherwise it yells at us with "a partial chunk must be sent in an
 	// individual packet" or just complains about a corrupted block.
-	alignment := outboundChunkSize - (int(s.offset) % outboundChunkSize)
+	alignment := (outboundChunkSize - (int(s.offset) % outboundChunkSize)) % outboundChunkSize
 	if alignment > 0 && alignment < packetLength {
 		packetLength = alignment
 	}
--- a/rpc/block_writer_test.go
+++ b/rpc/block_writer_test.go
@@ -99,3 +99,24 @@ func TestWriteFailsOver(t *testing.T) {
 	assert.EqualValues(t, 1048576, n)
 	assert.EqualValues(t, 0xb35a6a0e, hash.Sum32())
 }
+
+func TestPacketAlignment(t *testing.T) {
+	// Verify that alignment is correctly handled in blockWriteStream.
+	bws := &blockWriteStream{}
+	// Write some data so we can test making packets.
+	bws.buf.Write(make([]byte, outboundPacketSize*3))
+
+	packet := bws.makePacket()
+	// if we are totally aligned, then the packet should be outboundPacketSize.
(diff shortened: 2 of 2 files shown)


adamfaulkner:
BTW I needed this in order for performance to be adequate. Uploading a 8GB file was taking > 1hr.

adamfaulkner:
I added an issue #84

rpc/block_writer_test.go
```diff
@@ -99,3 +99,24 @@ func TestWriteFailsOver(t *testing.T) {
...
+
+func TestPacketAlignment(t *testing.T) {
+	// Verify that alignment is correctly handled in blockWriteStream.
+	bws := &blockWriteStream{}
+	// Write some data so we can test making packets.
+	bws.buf.Write(make([]byte, outboundPacketSize*3))
+
+	packet := bws.makePacket()
+	// if we are totally aligned, then the packet should be outboundPacketSize.
+	if len(packet.data) != outboundPacketSize {
```
colinmarc: Can you use `assert.NotEqual` here (and `assert.Equal` below)?

rpc/block_writer_test.go
```diff
@@ -99,3 +99,24 @@ func TestWriteFailsOver(t *testing.T) {
 	assert.EqualValues(t, 1048576, n)
 	assert.EqualValues(t, 0xb35a6a0e, hash.Sum32())
 }
+
+func TestPacketAlignment(t *testing.T) {
+	// Verify that alignment is correctly handled in blockWriteStream.
+	bws := &blockWriteStream{}
+	// Write some data so we can test making packets.
```
colinmarc: You can get rid of these two comments, they're just noisy.

rpc/block_write_stream.go
```diff
@@ -173,7 +173,7 @@ func (s *blockWriteStream) makePacket() outboundPacket {
 	// gets unhappy unless we first align to a chunk boundary with a small packet.
 	// Otherwise it yells at us with "a partial chunk must be sent in an
 	// individual packet" or just complains about a corrupted block.
-	alignment := outboundChunkSize - (int(s.offset) % outboundChunkSize)
+	alignment := (outboundChunkSize - (int(s.offset) % outboundChunkSize)) % outboundChunkSize
```
colinmarc: I think I see the problem now, ugh. An extra `%` just makes this line more confusing, though. I think something like this would scan better:

```
alignment := int(s.offset) % outboundChunkSize
if packetLength > outboundChunkSize && alignment != 0 {
    packetLength = outboundChunkSize - alignment
}
```

What do you think?

colinmarc:
Gah, thanks for catching this. Super terrible. I've got some nitpicks on the fix and the test changes, then I'll merge.

colinmarc:
I merged your fix in https://github.com/colinmarc/hdfs/commit/cdda132fb253decd0fb23f1bb183bad5fd8c7fe4. Thanks again!

adamfaulkner:
Ah sorry for not addressing those nitpicks, I've been super busy! Thanks for merging my fix.

PR #93, opened 2017-12-06
Expand range of completion

dlamblin:
Will complete commands in first position after command-name.
Completes hdfs files in second or third position for relevant commands.
Completes local files in second [for put] or third [for get, getmerge] position for relevant commands.

### DIFF
--- a/cmd/hdfs/bash_completion
+++ b/cmd/hdfs/bash_completion
@@ -1,20 +1,61 @@
 #!/bin/bash
 
+#Valid commands:
+#  ls [-lah] [FILE]...
+#  rm [-rf] FILE...
+#  mv [-nT] SOURCE... DEST
+#  mkdir [-p] FILE...
+#  touch [-amc] FILE...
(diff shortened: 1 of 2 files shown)


colinmarc:
Hi @dlamblin  - thanks for the work on this!

Even after `brew install bash-completion`, I didn't get a `compopt` binary on my mac. So unfortunately this doesn't work out of the box for me.

colinmarc:
(Also, I fixed the protoc problem in travis in https://github.com/colinmarc/hdfs/commit/b11655925068d29130276ed7f12d153031120443)

dlamblin:
Oh that explains why Travis failed, probably.
On reflection I'd installed bash (v4+) from Brew for better completion support of ssh-config entries and git; probably compopt is a built in added in v4. Also I got the Brew package bash- completion2.
Maybe some Mac OS release will include bash v4+ in the future and you can revisit this then.

Appreciate the excellent work on your tool, I frequently find it handy.

Issue #44, opened 2016-06-30
Allow put from stdin


SuperQ:
It would be useful to be able to put from stdin.

For example:

```
tar -cz path/ | hdfs put - /path/tar.gz
```

The way `cmd/hdfs/put.go` is structured makes this a little tricky, but I think I can refactor this.

colinmarc:
Ah, good idea!

fdv:
👍 
I'd love to have it too, it's a killing feature for me.

PR #98, opened 2018-01-23
put can read from stdin by specifying the '-' source file

Shastick:
Here is a small proposal for `put` ing via `STDIN`, as requested in #44.

I'm relatively new to GO (and the github workflow as well...) and more than happy to read any comment!


colinmarc:
Hi @Shastick,

Thanks for the PR! This looks like a reasonable approach, but can you add some tests and fix the build, to start? Also, please run `goimports -w` to format your code.

colinmarc:
(If you rebase, the protoc problem in travis should be fixed)

Shastick:
Thanks for the feedback. Will do this asap :)

Shastick:
Hoping the test is enough: I did not test scenarios where it should fail (existing file/directory with the same name)

Shastick:
Added test cases for expected failures

cmd/hdfs/put.go
```diff
@@ -29,6 +31,48 @@ func put(args []string) {
...
+}
+
+func putFromStdIn(client *hdfs.Client, dest string) {
+	// If the destination exists, regardless of what it is, bail out.
+	existing, err := client.Stat(dest)
+	if err == nil {
+		if existing.IsDir() {
+			fatal(fmt.Errorf("can't write from STDIN into a directory: %s", dest))
+		} else {
+			fatal(fmt.Errorf("won't write to an existing file: %s", dest))
```
colinmarc: Both branches here can be just `fatal(&os.PathError{"put", dest, os.ErrExist})`

cmd/hdfs/put.go
```diff
@@ -29,6 +31,48 @@ func put(args []string) {
...
 	}
 
+	if filepath.Base(source) == "-" {
+		putFromStdIn(client, dest)
+	} else {
+		putFromFile(client, source, dest)
+	}
+}
+
+func putFromStdIn(client *hdfs.Client, dest string) {
```
colinmarc: Nitpick: this should be `putFromStdin` (without the `I` capitalized), since the standard library spells it that way.

cmd/hdfs/test/put.bats
```diff
@@ -64,6 +64,33 @@ mkdir /_test_cmd/put/existing.txt: file already exists
...
+  assert_success
+
+  run bash -c "$HDFS cat /_test_cmd/put_stdin/mobydick_stdin.txt > $BATS_TMPDIR/mobydick_stdin_test.txt"
+  assert_success
+
+  SHA=`shasum < $ROOT_TEST_DIR/test/mobydick.txt | awk '{ print $1 }'`
+  assert_equal $SHA `shasum < $BATS_TMPDIR/mobydick_stdin_test.txt | awk '{ print $1 }'`
+}
+
+@test "put from stdin into existing file" {
```
colinmarc: How does the normal `hadoop -fs` behave in this case, out of curiosity?

cmd/hdfs/put.go
```diff
@@ -29,6 +31,48 @@ func put(args []string) {
...
+			fatal(fmt.Errorf("can't write from STDIN into a directory: %s", dest))
+		} else {
+			fatal(fmt.Errorf("won't write to an existing file: %s", dest))
+		}
+	} else if !os.IsNotExist(err) {
+		fatal(err)
+	}
+
+	mode := 0755 | os.ModeDir
+	// make sure parent dir exists
```
colinmarc: This and the next few comments are unnecessary. Also, for future reference, comments should be in complete sentences.

cmd/hdfs/test/put.bats
Shastick: Just had a try: `put: `/tmp/testme': File exists`

Shastick:
The failure seems unrelated (only happened on one of the 4 runs) for a rm test:

```
not ok 50 rm dir without -r, but with -f
# (from function `setup' in test file cmd/hdfs/test/rm.bats, line 8)
#   `$HDFS touch /_test_cmd/rm/b' failed
# no available namenodes: dial tcp 127.0.0.1:9000: i/o timeout```

colinmarc:
Merged in `47855e4`. Thanks so much for the contributions!

PR #99, opened 2018-01-30
Support for kerberized hadoop clusters.

Shastick:
Notes:
  - requires a patch to the krb5 lib to avoid setting a subkey when generating the authenticator.
  - only supports a QOP of "authentication". Other modes are not supported.


Shastick:
Note: I mostly used log.Panic(error) whenever something went wrong for bubbling up information fo debugging. I guess it would be better to return errors in certain places (ie, in sasl.go?) but more or less Ok from the main.go file?

colinmarc:
I'm going to leave some comments as I read through, partly just to keep track for myself - but no pressure on those.

Shastick:
Looks like there is an issue with the dependencies on Travis. Not sure to understand where the problem lies: are the deps in files living in /vendor not pulled automatically?

rpc/namenode.go
```diff
@@ -310,21 +318,38 @@ func (c *NamenodeConnection) readResponse(method string, resp proto.Message) err
...
 // |  varint length + IpcConnectionContextProto                |
 // +-----------------------------------------------------------+
 func (c *NamenodeConnection) writeNamenodeHandshake() error {
+	if c.kerberosClient != nil {
+		return c.writeNamenodeKerberosHandshake()
+	}
+	return c.writeNamenodeNoAuthHandshake()
+}
+
+func (c *NamenodeConnection) writeNamenodeNoAuthHandshake() error {
```
colinmarc: Doesn't really seem necessary to make this two separate functions. There's only two places where it branches that I see: choosing which auth protocol flag to include and the call to `doKerberosHandshake`.

rpc/sasl.go
```diff
@@ -0,0 +1,221 @@
...
+	"gopkg.in/jcmturner/gokrb5.v3/iana/keyusage"
+	"gopkg.in/jcmturner/gokrb5.v3/types"
+)
+
+const saslRpcCallId int32 = -33
+
+// Run a Kerberos handshake with an hadoop Namenode server that is expected to live on the other side
+// of the connection.
+// Assumes an HRPC header has already been sent.
+func doKerberosHandshake(c *NamenodeConnection) {
```
colinmarc: This should definitely be returning the error, not panicking.

colinmarc:
No, dependencies should be committed directly into `vendor` as part of the PR. It looks like you got the krb code, but not the transitive dependencies.

colinmarc:
(Also, looks like `vendor.json` wasn't updated. Are you using the `govendor` tool?)

cmd/hdfs/main.go
```diff
@@ -169,24 +188,169 @@ func fatalWithUsage(msg ...interface{}) {
...
+	}
+
+	cl, clE := client.NewClientFromCCache(cc)
+	if clE != nil {
+		log.Panic(clE)
+	}
+
+	cl.WithConfig(cfg)
+	// TODO Config flag or whatever for people not using AD
+	cl.GoKrb5Conf.DisablePAFXFast = true
```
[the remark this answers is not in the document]
Shastick: This is strictly related to Kerberos, not to hadoop. No idea of what it does exactly, just that Active Directory may not support it (from https://github.com/jcmturner/gokrb5#active-directory-kdc-and-fast-negotiation )

conf.go
```diff
@@ -34,7 +34,8 @@ func LoadHadoopConf(inputPath string) HadoopConf {
 	var tryPaths []string
 
 	if inputPath != "" {
-		tryPaths = append(tryPaths, inputPath)
+		tryPaths = append(tryPaths, filepath.Join(inputPath, "core-site.xml"))
```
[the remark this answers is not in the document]
Shastick: Yep, the original failed to load anything when a directory was passed. If I remember well nothing was calling this with an input path different from "".

Shastick:
Nope, did not use govendor: just dropping the code into /vendor did the trick, though it probably worked because I still had the required dependencies. 

A coworker did not complain about missing things, but I don't know how exactly he built the code.

colinmarc:
Ok, I switched to `dep` in a127fd59f03d61560fb7811f4dee974fbd5e53fb, so you should be able to just run `dep ensure -add gopkg.in/jcmturner/gokrb5.v3` and it should do the right thing.

conf.go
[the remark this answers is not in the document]
colinmarc: Fixed (and tested) in https://github.com/colinmarc/hdfs/commit/01da9262e4bc9d976bb06b1ffb88554350224bf4

Shastick:
- So, for now I added all the jars needed for kerby (I found no place where to download them all). We can always build them otherwise.
 - starting the KDC and generating the keytabs based on an env var seems fine
 - I'm still wondering how to best manage the multiple hadoop configurations. Trying to set HADOOP_CONF_DIR when kerberos is enabled

Shastick:
Almost there: still need to figure out how to tell Hadoop to accept Kerberos. Any idea how to override the default configuration? I tried setting HADOOP_CONF_DIR with a core-size.xml containing settings requiring Kerberos to no avail.

colinmarc:
I think the problem is that you're setting it inside the script, and it doesn't get exported. Can you try setting it in `travis.yml`?

colinmarc:
I was doing some research on how other projects stub kerberos in tests. [This](https://github.com/requests/requests-kerberos/blob/master/.travis.yml) ([with accompanying script](https://github.com/requests/requests-kerberos/blob/master/.travis.sh)) looks promising.

Shastick:
Okay, Hadoop is now running with Kerberos and the local KDC. Now I need to tell the java HDFS client how to talk with it.

Shastick:
It's clunky, but I'm getting there.

https://travis-ci.org/colinmarc/hdfs/builds/336189988 shows a partial success (and a weird failure on the first one: only a checksum check failed due to a reason I'm not sure to understand)

I ignored the go tests for now to at least get an environment where the binary can be tested. 

I'll try to see how we can let the go unit tests run agains kerberos, but it might be a bit complicated or not possible for the tests regarding permission checks. (I had to fiddle with the permission settings of hadoop -> currently permissions are ignored when kerberos is enabled).

colinmarc:
Good work! The warning is probably messing with the checksum output.

Shastick:
I fixed the warning, still failing:
```
# expected: 
# 27c076e4987344253650d3335a5d08ce /_test/foo.txt
# actual:   27c076e4987344253650d3335a5d08ce /_test/foo.txt
```
Looks like there is a missing newline: the hashes are Ok. The really funny thing is that the tests with the hdp hadoop distribution are fine.

Makefile
```diff
@@ -26,9 +26,16 @@ install: get-deps
 	go install ./...
 
 test: hdfs
+# Run unit tests only without Kerberos.
+ifndef KERBEROS
```
colinmarc: Why skip all the unit tests? We can skip just the permission ones if necessary (with `t.Skip("skipping permission tests under kerberos")`)

README.md
```diff
@@ -98,6 +106,19 @@ override this in your `.bashrc` or `.profile`:
 
     export HADOOP_USER_NAME=username
 
+
+Kerberos support
+----------------
+Authentication via Kerberos (and authentication only) is supported.
+
+The binary will check the default locations for your kerberos and hadoop configurations. These can be overridden via environment variables `HADOOP_KRB_CONF`, and `HADOOP_CONF_DIR`.
```
colinmarc: The library does this now too, right?

client.go
```diff
@@ -37,13 +42,23 @@ func Username() (string, error) {
 	return currentUser.Username, nil
 }
 
+// If the kerberos client is set, returns its principal.
+// Otherwise, defers to Username()
+func username(krb5Client *krb.Client) (string, error) {
```
colinmarc: Call this `krbUsername` (and the comment should be something like `krbUsername returns the principle of krbClient, or the result of calling Username if krbClient is nil`)

colinmarc: But also rather than the passthrough, I'd just have a branching if statement below

kerberos.go
```diff
@@ -0,0 +1,137 @@
...
+	"gopkg.in/jcmturner/gokrb5.v3/client"
+	"gopkg.in/jcmturner/gokrb5.v3/config"
+	"gopkg.in/jcmturner/gokrb5.v3/credentials"
+	"gopkg.in/jcmturner/gokrb5.v3/keytab"
+)
+
+const HdfsDefaultServiceName = "nn"
+const KrbDefaultCfgPath = "/etc/krb5.conf"
+const HadoopAuthCfgPath = "hadoop.security.authentication"
+const HdfsDefaultConfDir = "/etc/hadoop/conf"
```
colinmarc: This is already defined in `conf.go`

README.md
Shastick: Currently building the kerberos client and setting it in the options still happens in `main.go`.

`client.go` has a few `NewX` functions: not sure if they should all check for Kerberos? Or just `New()`

OTOH, library users who need Kerberos support could build their options struct: I have no opinion on this but don't want to mess things up in `client.go` too much :)

kerberos.go
```diff
@@ -0,0 +1,137 @@
...
+
+// GetKrbClientIfRequired returns a kerberos client if the hadoop
+// configuration or the environment variables suggest one is required
+func GetKrbClientIfRequired(conf HadoopConf) *client.Client {
```
colinmarc: I think better would be to create a generalized `SetConnectionOptions` method on `HadoopConf`. Something like

```
// SetConnectionOptions sets the necessary namenode address(es) and authentication settings on clientOptions from the loaded configuration, or returns an error if not enough information is present in the configuration to do so.
func (conf *HadoopConf) SetConnectionOptions(clientOptions ClientOptions) error { ... }
```

rpc/namenode.go
```diff
@@ -35,13 +37,22 @@ type NamenodeConnection struct {
...
+	// ServicePrincipalName the service part of the SPN (<SERVICE>/<FQDN>, ie, nn/localhost) if Kerberos is enabled
+	servicePrincipalName string
 }
 
 // NamenodeConnectionOptions represents the configurable options available
 // for a NamenodeConnection.
 type NamenodeConnectionOptions struct {
 	Addresses []string
-	User      string
+	// User will contain the full user principal if kerberos is used
```
colinmarc: `User should...`

colinmarc:
I think including the jars is not really going to work. Can you take a look at the apt-based approach I linked above and see if we can crib it for this?

kerberos.go
Shastick: Agreed. Though, would it not make more sense to call this method on the ClientOptions structe? ie, have:

```
func (opts *ClientOptions) SetConnectionOptions(conf *HadoopConf)
```

(I'm a bit bothered by having a `Set` function on something that sets something in the parameter, but if that's a typical way to do it in Go why not :))

setup_kdc.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
Shastick: We could probably make `setup_kdc.sh` work locally more or less out of the box. Or move it to somewhere else?

kerberos.go
colinmarc: You're right, that's better.

colinmarc:
Can you change the [settings](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) on the PR to let me edit it? That way I can fuss with the code layout and documentation myself, rather than endlessly nitpicking in comments :)

Shastick:
If including the jars is not an option, we can also build them from Kerby's sources (and cache the result? It looks like travis allows this).

The apt options looks promising and might well end up being less messy than the current setup, but I won't have the time to dig into this now.

Shastick:
Jeez. Permission problems all over again :X could be linked to the fact that the 'nn' user does not exist.

Shastick:
Ok, after a night of sleep things are a bit clearer. Current situation:
  - With Kerberos enabled, appending to blocks does not seem to work, as well as some failover tests, and some calls to `chown`.
  - The command line tests seem to work with and without Kerberos (though the binary is not tested on travis due to the unit tests failing
  - With Kerberos disabled everything is green again.

Shastick:
Last thing to be done: get rid of the kerby jars. Two options:

  - check the suggested packaged kerberos docker image
  - build the jars from kerby's source
(https://github.com/apache/directory-kerby or http://directory.apache.org/kerby/download/download-sources.html)

This PR has already eaten up quite some time, so I'd be in favor of staying with the jars (either in the repo or building them) until I have more time or someone with more experience in stubbing Kerberos takes a look.

mxk1235:
awesome contribution!  any idea when it might get merged?

Shastick:
The test setup is still a bit rough (some jars are stored along in the repo). I'll see if I can get them to be built as part of the test process.

colinmarc:
Hey @Shastick,

I'd rather we not continue down the road with the jars. If it's ok with you, I'm going to pull this into an internal branch this week and see if I can crib that python kerberos travis setup. Then we'll keep it as a branch until a few more people have gotten a chance to try it in prod. How does that sound?

Shastick:
Sounds good to me :) 

I've given a try at the docker setup you linked to before yesterday. Current status:
  - KDC is started, principals can be added
  - interaction with the KDC is possible 
   
**current blocker**: hadoop fails to login to the KDC, most probably because of hostname validation (from the KDC's point of view the hadoop hostname isn't `localhost`).

Shastick:
So, it looks like running hadoop from within the same docker image will be the easiest way out. 

I'll see if I can give it a try later today.

Shastick:
@colinmarc I had a look. Besides running everything in a single docker container, there is also the option to set the current hostname in the namenode principal (as defined in the core-site file) when preparing the files. This has the strange effect of requiring some other properties in hdfs-site (and to run hadoop with sudo rights...):

```
     <property>
        <name>dfs.datanode.address:1002</name>
        <value>localhost</value>
    </property>
    <property>
        <name>dfs.datanode.hostname</name>
        <value>localhost</value>
    </property>
    <property>
        <name>dfs.datanode.http.address</name>
        <value>localhost:1003</value>
    </property>
```

The explicit hostname + sudo option is a bit messy. oing things in docker requires kind of an overhaul of the test process. 

What do you prefer?

I'll update the PR with something not too messy as a good base for you to have a look at.

rpc/namenode.go
```diff
@@ -309,27 +321,49 @@ func (c *NamenodeConnection) readResponse(method string, resp proto.Message) err
...
+		rpcVersion,
+		serviceClass,
+		authProto,
+	}
+
+	c.conn.Write(rpcHeader)
+
+	if authProto == saslAuthProtocol {
+		doKerberosHandshake(c)
 	}
```
staticmukesh: Kindly handle error condition here. Even in case of failed kerberos authentication, it doesn't return here and leads to `FATAL_INVALID_RPC_HEADER`. Kindly fix the issue.

kerberos.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if krb5Cfg == "" {
+		krb5Cfg = KrbDefaultCfgPath
+	}
+
+	// Now check if the credential cache or the keytab have been manually specified
+	keytabPath := os.Getenv("HADOOP_KEYTAB")
+
+	if keytabPath != "" {
+		return getKrbClientWithKeytab(krb5Cfg, keytabPath)
```
staticmukesh: Can we pass `keytabPath` to the function rather than relying only on environment variable ?

rpc/namenode.go
Shastick: Thanks, missed that one. It's obviously a mistake :)

kerberos.go
Shastick: Definitely OK with the idea, but it's slightly outside of the scope of this PR -> I'd rather wait until the branch is merged and make the hdfs library more user-friendly afterwards.

Shastick:
@colinmarc did you have any luck playing with docker?

mxk1235:
i just went through a similar exercise testing kerberized containers.  it's pretty difficult, especially with network names needing to match credentials.  i could possibly assist you in the interest of merging this change.

Shastick:
@mxk1235 assistance would be of great help: in your case, did you do everything within docker containers, or did you manage to fix the hostname problem?

If you have a public sample, I'd be happy to take a look at it.

mxk1235:
i was working with a docker image that contained kdc, zookeeper, and kafka.  kafka would auth against zk (locally), and clients would auth to kafka (externally).  kdc port needed to be exposed.

i setup the principals with a host.example.com domain, and built the image such that /etc/hosts that would resolve that dns name to itself.  on the client, i had to associate host.example.com to the kdc/zk/kafka container ip similarly.  all connection urls then had to use host.example.com as well.  you'll have to make sure that the principal.

i believe in your case you could get away with a docker image that contained the kdc/nn/dn, and reuse host.example.com everywhere like i did.

i really hope this helps.

Shastick:
Thanks for the input. Fiddling with /etc/hosts is definitely a good idea.

Regarding having the namenode in docker: it's probably the way to go. We could get there easily by mounting the jars into the container and running them from there.

melan:
I have a full KDC, Hdfs, zookeeper, HBase and few other services as a single docker compose with certs and keytabs enabled. Will it help you guys?

melan:
@colinmarc try this: https://github.com/melan/local_hbase 🤞 - you'll need to generate keytabs and certificates. Both of the scripts for that are in the `./scripts` folder. If you need a keytab for your test - add another line into the generate-keytabs.sh script.
There is a problem with keytabs in hbase, but I don't think you need it for this change.

mxk1235:
any progress on this?

Shastick:
@mxk1235 definitely still on my mind, I just wasn't able to take the time for it :(

hsawhney09:
@Shastick This is awesome. I am just wondering what is left to do in order to merge this? How can I help (I don't have much domain knowledge here though)

Shastick:
Basically: unit tests must pass for both a non-kerberized and a kerberized cluster. The integration tests already work in both cases.

Apart from that, the code works. 

If I recall correctly, the path the getting this to work is to run the whole test-suite to run within docker (ie, having a dockerized kerberos server, Namenode and hdfs client) so the hostnames don't get messed up for the Kerberos principals.

staticmukesh:
Hello @Shastick and @colinmarc, I have fixed build related issues to this PR. Since I can't modify this pull request, I have raised the PR at source [PR-16](https://github.com/Sqooba/hdfs/pull/16). 

Please review the PR and lets merge it asap.

Shastick:
Cherry-picked from @staticmukesh's PR. It's Christmas before time :)

didip:
Hi all, thank you for all of your hard work. I am curious when will the PR be merged.

Shastick:
@didip I'll wait a few more days (@colinmarc might be on vacation :)). If nothing has happened until then (probably end of next week) I'll do a little cleanup of our forked repo so it can easily be used until this is merged.

colinmarc:
I am on vacation :) My plan when I get back is to pull this into an internal branch and work on it there.

colinmarc:
I've pushed a WIP to https://github.com/colinmarc/hdfs/pull/133. If anyone is currently using this branch, I'd really appreciate if you switched to that one to test it out (and let me know if anything doesn't work).

colinmarc:
(as this moved to an internal branch at #133, going to close this)

Issue #100, opened 2018-02-12
recursive `ls`, ala `ls -R`


Shastick:
It would be nice to be able to list directories recursively, ala `ls -lR`.

I'm currently patching the library to be able to return a list of files living beneath a given root, and it looks like a recursive ls could be obtained easily (both for the CLI and the lib') by moving the `walk` function found in `paths.go` to somewhere reachable in the library.

colinmarc:
With #101, this should now be possible.

PR #101, opened 2018-02-13
Move walk() to the library, make it behave exactly like filepath's walk.

Shastick:
As a possible stepping stone to #100, and to allow library users to walk an hdfs path hierarchy as they would on their local file system through `filepath.Walk()`


colinmarc:
Thanks for the contribution! Merged in 7c31e90d217d46e87963cb6932b261d0c67640d0.

PR #114, opened 2018-05-10
Lazily create namenode connection

afking:
This allows the NamenodeConnection to be created without performing the initial handshake.

### DIFF
--- a/rpc/namenode.go
+++ b/rpc/namenode.go
@@ -42,6 +42,7 @@ type NamenodeConnection struct {
 type NamenodeConnectionOptions struct {
 	Addresses []string
 	User      string
+	Lazy      bool
 }
 
 // NamenodeError represents an interepreted error from the Namenode, including
@@ -87,7 +88,7 @@ func NewNamenodeConnection(address string, user string) (*NamenodeConnection, er
 }
 
 // NewNamenodeConnectionWithOptions creates a new connection to a namenode with
-// the given options and performs an initial handshake.
+// the given options and optionally performs an initial handshake.
 func NewNamenodeConnectionWithOptions(options NamenodeConnectionOptions) (*NamenodeConnection, error) {
 	// Build the list of hosts to be used for failover.
 	hostList := make([]*namenodeHost, len(options.Addresses))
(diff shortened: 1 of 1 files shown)


colinmarc:
Hi @afking - thanks for the suggestion! Can you talk a bit about the rationale here? I'd rather change the behavior to always be lazy than have an option, but I don't really see the advantage to either.

afking:
Hi @colinmarc, for our services we want to always have the service come up. This is so you can handle the errors on requests rather then setup like how sql.DB connections work. I would be happy if there is another way to solve this, tried to do the smallest changeset.

colinmarc:
I think changing the behavior would be too drastic, and the option is a bit unnecessary. Does fixing #120 solve your problem, since it can just retry?

afking:
@colinmarc on errors the struct is always returned nil. I couldn't fine a way to expose the underlying conn without it doing the connection. Could we maybe get rid of testing the connection on all initialization instead of the option?

colinmarc:
Hi @afking, does #139 fix this for you? Or alternatively,

> I couldn't fine a way to expose the underlying conn without it doing the connection. 

Why do you need the underlying struct but disconnected? Couldn't you just have a

```go
struct lazyHDFSClient {
   client *hdfs.Client
   options hdfs.Options
}

func (l lazyHDFSClient) Get() (*hdfs.Client, error) {
  if l.client != nil {
    return l.client, nil
  }

  l.client, err = hdfs.NewClient(l.options)
  return l.client, err
}```

Then instead of constructing the `Client` early, you could just construct one of those?

**Comment by afking** [comment]
Could use a constructor but then each call to HDFS becomes rather verbose:
```
client, err := l.Get()
if err != nil
   return err
}

resp, err := client.DoRequest(...)
if err != nil {
    return err
}
```
What about wrapping the handshake in a `Ping()` function and just validating the arguments on creation: https://golang.org/pkg/database/sql/#Open ?

colinmarc:
> What about wrapping the handshake in a Ping() function and just validating the arguments on creation

In my opinion, that's a particularly ugly bit of interface papier-mâché necessitated by the differences in sql driver implementations; I don't think we have the same requirement here. 

Go is already pretty verbose so I think what you wrote looks pretty ok :) I'm going to close this, but feel free to make more suggestions if you have any.

PR #115, opened 2018-05-11
Fix FileReader ReadAt errors

afking:
Using `io.ReadFull()` changes the error message to `io.ErrUnexpectedEOF` instead of the expected `io.EOF`. This fixes the behavior to be handled like `os.FIle` https://golang.org/pkg/os/#File.ReadAt

### DIFF
--- a/file_reader.go
+++ b/file_reader.go
@@ -192,12 +192,16 @@ func (f *FileReader) ReadAt(b []byte, off int64) (int, error) {
 		return 0, io.ErrClosedPipe
 	}
 
+	if off < 0 {
+		return 0, &os.PathError{"readat", f.name, errors.New("negative offset")}
+	}
+
 	_, err := f.Seek(off, 0)
 	if err != nil {
 		return 0, err
 	}
 
-	return io.ReadFull(f, b)
+	return f.Read(b)
 }
 
 // Readdir reads the contents of the directory associated with file and returns


colinmarc:
Good call! Tested with this code:

```
package main

import "os"

func main() {
	// foo.txt contains "foobar"
	f, err := os.Open("foo.txt")

	if err != nil {
		panic(err)
	}

	b := make([]byte, 10)
	_, err = f.ReadAt(b, 3)
	panic(err)
}
```

colinmarc:
Needed to make some tweaks - this change returned `nil`, not `io.EOF`, and we want the behavior of `io.ReadFull` where it calls `Read` in a loop.

afking:
@colinmarc awesome thanks, bit confused why the Read didn't work? Should this not behave like Read() with a seek method?

afking:
https://golang.org/pkg/os/#File.Read I think `Read()` should return `io.EOF`.

colinmarc:
@afking `File.ReadAt` actually calls `Read` in a loop, just like `ReadFull`:

```
for len(b) > 0 {
  		m, e := f.pread(b, off)
  		if e != nil {
  			err = f.wrapErr("read", e)
  			break
  		}
  		n += m
  		b = b[m:]
  		off += int64(m)
  	}
```

And `Read` doesn't have to return `io.EOF` on a partial read - it can just return `n < len(b)` and `nil`.

colinmarc:
> https://golang.org/pkg/os/#File.Read I think Read() should return io.EOF.

It doesn't until the second call to `Read`. Try running this code:

```
package main

import "os"
import "log"

func main() {
	// foo.txt contains "foobar"
	f, err := os.Open("foo.txt")

	if err != nil {
		panic(err)
	}

	b := make([]byte, 10)
	n, err := f.Read(b)
	log.Fatal(n, err)
}
```

I get `7 <nil>`.

afking:
@colinmarc yep it does 👍 thanks for taking the time to clarify.

PR #116, opened 2018-05-20
Add Viewfs feature support

kirbyzhou:
Add very limited viewfs feature support.
sample conf files are put in test/conf-view folder.
sample test cast are put in cmd/hdfs/test/viewfs.bats

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ hdfs: clean $(SOURCES)
 install: get-deps
 	go install ./...
 
+check: test
 test: hdfs
 	go test -v -race $(shell go list ./... | grep -v vendor)
 	bats ./cmd/hdfs/test/*.bats
@@ -38,4 +39,4 @@ release: hdfs
 	cp hdfs README.md LICENSE.txt cmd/hdfs/bash_completion $(RELEASE_NAME)/
 	tar -cvzf $(RELEASE_NAME).tar.gz $(RELEASE_NAME)
 
-.PHONY: clean clean-protos install test release
+.PHONY: clean clean-protos install check test release
--- a/client.go
+++ b/client.go
@@ -1,27 +1,54 @@
 package hdfs
 
 import (
-	"io"
-	"io/ioutil"
 	"os"
 	"os/user"
-
-	hdfs "github.com/colinmarc/hdfs/protocol/hadoop_hdfs"
-	"github.com/colinmarc/hdfs/rpc"
+	"path/filepath"
+	"time"
 )
 
-// A Client represents a connection to an HDFS cluster
-type Client struct {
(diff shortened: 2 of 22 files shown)


colinmarc:
Hi @kirbyzhou,

Thanks so much for the work on this! I think it's a bit more invasive than I'm ok with, given the relative unimportance of the feature to most people. Is there a way this could be accomplished just by adding a `ViewFSClient` and not an interface/renaming the existing client/etc?

colinmarc:
Yeah, I think I'd rather add a `ViewFSClient` on its own and then either have that be unsupported by the command line or work in support there somehow.

The other option would be to build in support directly into `Client`. I don't have a good handle on what that diff would look like.

colinmarc:
Could we do the same thing in this PR with a map of `Client`s, but use a of `NamenodeConnection`s instead? I feel that wouldn't be very invasive - it would require adding a method to get the right `NamenodeConnection` for a given path, which would replace the `namenode` field in the `Client` struct. Then you could just use that in every method instead of `c.namenode`.

kirbyzhou:
About using a map of `NamenodeConnection`s instead.
Do you mean that: instead of adding a `ViewfsClient` as wrapper of `Client`s, but modify the code of all existing public method of `Client`. 

For example:

```
func (c *Client) mkdir(dirname string, perm os.FileMode, createParent bool) error {
	dirname = path.Clean(dirname)
        var nclient Namenode

	// get namenode connection and new path
	nclient, dirname = c.getNamenodeAndRewritePath(dirname)

	// .... 
	// .... 

	err = nclient.Execute("mkdirs", req, resp)
	if err != nil {
		// ....
	}

	return nil
}
```

I think it would works, but it may makes problems later.
Hadoop support hdfs://, file://, viewfs://, maybe nfs:// now,  and Viewfs is not restricted on hdfs://. If we want to extend it to accept file:// or other FS in the future, an abstraction layer is needed to express the client of any file system instead of embedded `Namenode`s. I wanna the old `Client` keeps simple and contacting with only one exact Namenode.

Something more, high-available and round-robin of Namenode can be implemented with the same abstraction layer mechanism.

Maybe we can change the name 'SimpleClient' back to Client, add new interface such as IClient, provides new API to smartly return ViewfsClient or Client as IClient, and change the code of cmd/hdfs to take the advantage of new API.

colinmarc:
Hm, fair point. At this point I'm leaning towards the idea that viewfs should be its own standalone library, though. I definitely have no interest in adding in support for `file://`, etc.

colinmarc:
Hi @kirbyzhou, 

I've thought about it for a while, and I don't think I'd like to expand the scope of the library this much. So I'm going to close this. But I really appreciate the contribution!

PR #117, opened 2018-05-22
Reads on nil buffers

afking:
Fixes read behavior when passing in 0 length buffers.


colinmarc:
Ha, good catch. How did you run into this?

afking:
@colinmarc we sync io.Pipe reads with nil buffers.

PR #123, opened 2018-06-13
use HostName instead of IpAddr for multi-homed HDFS

hollow:
In multi-homed HDFS setups the IP address is not accessible from all nodes and the hostname should be used anyway. It's not the client libraries job to workaround broken DNS/resolver setups.


colinmarc:
This is definitely right, so I'm going to merge, but I'm worried about compatibility, so I'll do a quick version bump.

Itfly:
@colinmarc I got the issue because of using HostName which is not resolved to ip address in my company, any workaround for this issue?

PR #125, opened 2018-06-22
the result of filepath.Join doesn't work if the client runs on Windows

elebore:
the result of filepath.Join doesn't work if the client runs on Windows while the server runs on Linux


colinmarc:
Great catch, thanks!

Issue #76, opened 2017-04-12
Timeouts are not configurable for RPC 


peay:
The timeouts for RPC do not seem to be configurable. From `rpc.go`:

```.go
var (
	connectTimeout  = 1 * time.Second
	namenodeTimeout = 3 * time.Second
	datanodeTimeout = 3 * time.Second
)
```

These values are rather low (especially the connection timeout), and I have been experiencing quite a lot of timeouts.

Being able to configure these would be very helpful. This does not have to be per client -- a global setting would be fine too.

colinmarc:
Hi @peay,

Good point - this should definitely be configurable. This dovetails well with the need for a configurable dialer in #70.

colinmarc:
With #77, there now exists a ClientOptions struct, so this should now be implementable.

Itfly:
@colinmarc  Any update for this issue, i also got error: dial tcp: i/o timeout

xxh2000:
@I meet this problem too, How do you reslove this? thanks.

xxh2000:
@Itfly  I meet this problem too, How do you reslove this? thanks.

Itfly:
@colinmarc Yes it works after i adding the datanode host to our company's DNSSelfService. However, i suggest to make it configurable about using hostname or ip address (https://github.com/colinmarc/hdfs/pull/123)

PR #130, opened 2018-07-10
Global rpc.ConnectTimeout set via client options.

curt-hash:
This addresses the lack of a configurable connect timeout (see https://github.com/colinmarc/hdfs/issues/76).

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/user"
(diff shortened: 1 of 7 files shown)


colinmarc:
Hi @curt-hash, thank you for the contribution, and I'm really sorry for the duplicated work, but I decided to go a slightly different direction with this (in #139). Cheers, and thanks again.

PR #132, opened 2018-07-13
info may be nil

elebore:
info may be nil caused by network error or other errors


colinmarc:
Good catch! Thanks so much for the contribution. 🤓

PR #137, opened 2018-07-29
Create RPC options, pass them through from the client

### LINKED ISSUE #118: Is there any configure function to add  dfs.client.use.datanode.hostname ?

barakmich:
This change adds the ability to pass client options through to the underlying RPC calls. This is useful for `dfs.client.*` options from hdfs-site.xml, including timeouts and the like.

It seemed right to pass in an RPC options struct, so I created one, and subsumed related parameters (eg, `clientName`) where it made sense. This provides a nice general framework for other such connection options (usually pertaining to connections made to datanodes).

Adds support for `dfs.client.use.datanode.hostname`, which fixes #118, and functionally reverts `a79e6ee`, while respecting `hdfs-site.xml` for those that use this functionality.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,6 +15,7 @@ import (
 type Client struct {
 	namenode *rpc.NamenodeConnection
 	defaults *hdfs.FsServerDefaultsProto
+	Options  ClientOptions
 }
 
 // ClientOptions represents the configurable options for a client.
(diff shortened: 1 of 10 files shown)


colinmarc:
Hi @barakmich! So sorry for the duplicated work - I actually was just about to push #138 when I got the email notification for this. I really appreciate the contribution.

I actually originally was going to take the approach of passing around a global config struct like you do here, but I decided to do timeouts with `Context` instead, which made it less necessary. Hopefully I don't come to regret that decision.

Issue #140, opened 2018-08-10
SSL support


mxk1235:
i don't see anything that relates to SSL encryption.  is this library compatible with hdfs clusters that setup the following properties?

```
  <property>
    <name>hadoop.ssl.enabled</name>
    <value>true</value>
  </property>
  <property>
    <name>hadoop.ssl.client.conf</name>
    <value>ssl-client.xml</value>
    <final>true</final>
  </property>
```

thanks in advance!

colinmarc:
Hi @mxk1235,

Are you actually running into problems with this? As I understand it, SSL affects only the HTTP API, but this library connects using the protobuf API.

I'm going to close this, but let me know if you have any issues using the library or commandline client on your cluster.

mxk1235:
thanks for the response.  yeah sounds right.  i was debugging connecting to a cluster and this config looked suspicious, but apparently shouldn't be an issue

mxk1235:
@colinmarc i wanted to follow up on hdfs data transfer encryption.  

besides hadoop.ssl properties on the namenode, there are also `hadoop.rpc.protection` setting in core-site.xml which can be set to `privacy`, as well as  `dfs.encrypt.data.transfer` which is `true` or `false` in hdfs-site.xml.   i believe those apply to the protobuf rpc interface, not http traffic.  

on first look this library just sets the default value for that property as `false`
https://github.com/colinmarc/hdfs/blob/f87e1d64bc48c85b07cab32d23c97788e885b31b/internal/protocol/hadoop_hdfs/hdfs.proto#L402
 and there is no way of creating a client with that property set to true. 
https://github.com/colinmarc/hdfs/blob/f87e1d64bc48c85b07cab32d23c97788e885b31b/client.go#L122

however, there is a fetchDefaults() function, that's only invoked by file_writer, but not by file_reader (e.g. Stat(), Readdir(), and Read() methods).

can you comment if i'm digging in the right place, and whether the encrypted part of the protocol applies to the read functionality?

mxk1235:
to be more precise, here's the relevant properties from core-site.xml

```
  <property>
    <name>hadoop.security.authentication</name>
    <value>kerberos</value>
  </property>
  <property>
    <name>hadoop.security.authorization</name>
    <value>true</value>
  </property>
  <property>
    <name>hadoop.rpc.protection</name>
    <value>privacy</value>
  </property>
```

and hdfs-site.xml

```
  <property>
[...]

PR #141, opened 2018-08-10
WIP: ssl/tls beginnings

mxk1235:
i was not able to build the head of master locally, but i hope to start some discussion around this, which looks like a simple enough change to add.


colinmarc:
Oof. I misspelled 'principal'. That hurts.

colinmarc:
As I wrote on #140, SSL should only affect the HTTP API, not the protobuf one. And I can't fix the spelling mistake because it would break backwards compatibility. I'll have to wait for v3 😢

PR #152, opened 2018-09-25
Fix race in fetchDefaults

efirs:


colinmarc:
Hi @efirs!

This is definitely not the only place `Client` is not thread-safe - generally speaking, it isn't advertised as such. You should use one client per thread if you need to do that.

Going to close this, but please tell me if you think I should reopen.

efirs:
This fixes the only reason why the client is not thread safe. My single client multi-threaded tests never indicate any other problems.

efirs:
I see that there is another attempt to fix the same issue https://github.com/colinmarc/hdfs/pull/181, which is closed as well.

@colinmarc can you give an example where client is not thread safe besides the place this PR fixes?

PR #154, opened 2018-09-25
Send heart beats to datanode

efirs:


colinmarc:
Hi @efirs! Thanks for this and the other contributions. I'll try to look at all of them today. For this one: what is the specific problem this solves? Is it possible to add a test case that breaks without this code?

efirs:
The test case is:
```
Write()
Sleep(1.5*Minute)
Write()
```
Second write fails without this fix.
The fix adds missing functionality from: https://apache.googlesource.com/hadoop/+/trunk/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DataStreamer.java

I've added the test case, but it increases test time by 1.5 minutes, the better one would lower HDFS idle timeout.

colinmarc:
Merged in #208. Sorry to make you wait more than a year!

PR #155, opened 2018-09-25
Reconnect to the same namenode after first write error

efirs:
Active namenode may close idle tcp connection, so next client write to
this connection will fail with "broken pipe" error.
But at this point there is no point in penalizing the namenode and try to
failover to other namenodes.
This change allows to try to reconnect to the same namenode after first write
error.

### DIFF
--- a/rpc/namenode.go
+++ b/rpc/namenode.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"os"
 	"sync"
 	"time"
 
@@ -72,6 +73,7 @@ type namenodeHost struct {
 	address     string
 	lastError   error
 	lastErrorAt time.Time
+	writeError  bool
 }
 
 // NewNamenodeConnection creates a new connection to a namenode and performs an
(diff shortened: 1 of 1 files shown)


efirs:
The test case is simple:
```
func TestFileReadDelayed(t *testing.T) {
       client := getClient(t)

       _, err := client.Open("/_test/foo.txt")
       require.NoError(t, err)

       time.Sleep(30 * time.Second)

       _, err = client.Open("/_test/foo.txt") // any request to namenode fails here
       require.NoError(t, err)
}
```
The number of connections before checking for idle should be lowered for test case to fail without the fix:
`echo "<configuration><property><name>ipc.client.idlethreshold</name><value>1</value></property></configuration>" >${HADOOP_HOME}/etc/hadoop/core-site.xml`. Full test case here: https://github.com/colinmarc/hdfs/commit/c21ad594f2496742aca3ef1799f6fd7596db3dd9

Without this fix it's impossible to reliably produce big data to HDFS.
With all the fixes I filed as PRs, we can reliable stream hundreds of TB of data.

PR #162, opened 2018-11-14
Make Client.Remove non-recursive, add Client.RemoveAll

twz123:
Fix an oversight in the implementation of Remove, which wasn't intended to be recursive. Also add a recursive version which is called RemoveAll.

This is a breaking API change that will not result in compilation errors.

~~Add a non recursive version of `Client.Remove()` which is named `Client.Delete()`.~~

~~I think the original method `Remove` should be called `RemoveAll`, and `Delete` in turn should be called `Remove`, but this would be definitely a cruel API change, so I just called it that way. Sorry for the not very clever name, suggestions welcome ;)~~

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -90,7 +90,7 @@ func touch(t *testing.T, path string) {
 func touchMask(t *testing.T, path string, mask os.FileMode) {
 	c := getClient(t)
 
-	err := c.Remove(path)
+	err := c.RemoveAll(path)
 	if err != nil && !os.IsNotExist(err) {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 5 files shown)


colinmarc:
I think this is just a bug. `Remove` should not be deleting recursively, since the documentation says it doesn't, and that the functions should match the `os` equivalents. I'm fine with the functionality change under a (minor) version bump.

Also, `RemoveAll` is apparently missing, probably due to the same oversight.

twz123:
Okay, adjusted PR accordingly.

colinmarc:
Merged in `6f7e441ec688730014b4ca08657db358d7a45b87`. Thanks so much for the contribution!

PR #171, opened 2019-02-23
Expose file ID

dtaniwaki:
I'm implementing [HDFS support in `argo-events`](https://github.com/argoproj/argo-events/issues/142), and need to check `File ID` to watch rename events. Could you consider exposing this value?

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -104,3 +104,8 @@ func (fi *FileInfo) OwnerGroup() string {
 func (fi *FileInfo) AccessTime() time.Time {
 	return time.Unix(int64(fi.status.GetAccessTime())/1000, 0)
 }
+
+// FileID returns the file ID which identifies the same file in HDFS
+func (fi *FileInfo) FileID() uint64 {
+	return fi.status.GetFileId()
+}


colinmarc:
@dtaniwaki hi! Thanks for the contribution. Does something like the following work for you?

```go
if status, ok := fi.Sys().(*hadoop_hdfs.HdfsFileStatusProto); ok {
    return status.GetFileID()
}```

`Sys` is part of the [`os.FileInfo` interface](https://godoc.org/os#FileInfo), but `FileID` is not.

Going to close this, but happy to hear arguments otherwise.

dtaniwaki:
I can’t import it because it’s under internal directory.

dtaniwaki:
@colinmarc Could you tell me how to import `hadoop_hdfs.HdfsFileStatusProto`?
As far as I know, internal packages can not be imported outside.
https://stackoverflow.com/questions/41571946/internal-packages-in-go

colinmarc:
Hm. That is indeed a problem. I guess I broke that when I moved it to internal.

I have to think about the right way to fix this, but for now you could copy the .proto file into your project 🙁

colinmarc:
With 322c085, you should be able to cast as `hdfs.FileStatus` instead. Sorry for the six-month delay!

PR #177, opened 2019-03-28
support multi storagetype

Loncoder:
I'm implementing BlockWriter to support block  be written to multi storage. Need to provide current storage type and target storage types , could you please  consider exposing these values?


colinmarc:
Hi @Loncoder,

Sorry for the delayed response? What is the use for these parameters? I'm not familiar with multi storage.

Loncoder:
Hi @colinmarc 
Glad to hear your respond, Apache Hadoop 2.3+ version begins to support heterogeneous storages. In hadoop document https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/ArchivalStorage.html , multi storage type can be chosen[MEMORY/SSD/DISK/ARCHIVE]. But now our go client can only write data to DISK(default) type. So if we consider this feature, we can apply our client to support high version of HDFS Service

colinmarc:
Hi @Loncoder,

Sorry for the long delay. I don't think I'm interested in supporting heterogenous storage in this project.

hsawhney09:
@colinmarc Can please revisit this decision of not supporting heterogeneous storage. It is becoming more and more common to have heterogenous storage types

colinmarc:
Hi @hsawhney09,

Thanks for bringing it up!

If you'd be willing to submit a PR or at least summarize what's required on the client side to support this feature, I would consider it.

Colin

PR #180, opened 2019-04-06

harshavardhana:
The scope of this PR is to only support user extended attributes,


harshavardhana:
ping @colinmarc

harshavardhana:
Closing this for now, keeping the work at the fork minio/hdfs

colinmarc:
HI @harshavardhana,

I'd happily accept this feature with tests (and using the official proto library instead of gogo/protobuf).

PR #181, opened 2019-04-19
make the client threadsafe

jhaynie:
We had a problem when multiple go routines are trying to use the client to push to HDFS with a race condition.


colinmarc:
Hi @jhaynie,

This is unfortunately not the only place that the client is un-thread-safe. I recommend using one client per goroutine.

PR #195, opened 2019-07-30
Implement touch [-am] features

xzy256:
`hdfs -h` display `touch` supports `-amc` argments, but it just realize `-c` function.  This PR focus to implement `-am` features .

### DIFF
--- a/cmd/hdfs/touch.go
+++ b/cmd/hdfs/touch.go
@@ -1,11 +1,12 @@
 package main
 
 import (
+	"github.com/colinmarc/hdfs/v2"
 	"os"
 	"time"
 )
 
-func touch(paths []string, noCreate bool) {
+func touch(paths []string, noCreate bool, accessTime bool, modifyTime bool) {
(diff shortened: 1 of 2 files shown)


colinmarc:
Hi @xzy256, could you please add some tests to this to `touch.bats`?

I fixed the docs in 13aa9393bb9244d3a47ebbf95a49d5f70998a787

cmd/hdfs/touch.go
```diff
@@ -25,18 +26,25 @@ func touch(paths []string, noCreate bool) {
...
 		if exists {
-			now := time.Now()
-			mtime := now
-			atime := now
-
-			err = client.Chtimes(p, mtime, atime)
+			if accessTime {
+				now := time.Now()
+				atime := now
+				err = client.Chtimes(p, atime, finfo.ModTime())
```
colinmarc: I think this error can be hidden in the case of `-am` (both flags). There should just be one `Chtimes` call at the end.

PR #198, opened 2019-08-29
added go.mod to build and use hadoopconf package properly

AnirudhVyas:
please review and add. You can reject this and add your own - whatever you wish, we need to be able to use hadoopconf package properly.

Thanks in advance!


AnirudhVyas:
something is wrong ... closing.

PR #206, opened 2019-11-20
Add snapshots to v1

devoxel:
I'm not sure if this repo is actively maintained, but in case it is, here's some features for snapshots.

This adds the following functions to :

- AllowSnapshot
- DisallowSnapshot
- CreateSnapshot
- DeleteSnapshot

The reasoning behind backporting to v1 is that not everybody has switched over to `go modules` and users of godep will have issues importing this library unless it's pinned to v1.

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -1,0 +1,72 @@
+package hdfs
+
+import (
+	hdfs "github.com/colinmarc/hdfs/protocol/hadoop_hdfs"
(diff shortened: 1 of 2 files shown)


colinmarc:
Hi Aaron,

While I'm happy to accept bug fixes to v1, I don't want to add any features there. I've merged your other PR, though.

Thanks again for the contribution.

PR #215, opened 2020-01-05
add explicit inheritance

flashpixx:
add explicit inheritance to make it compatible with io interfaces

### DIFF
--- a/file_reader.go
+++ b/file_reader.go
@@ -17,6 +17,8 @@ import (
 // io.Reader, io.ReaderAt, io.Seeker, and io.Closer, and can only be used for
 // reads. For writes, see FileWriter and Client.Create.
 type FileReader struct {
+	io.ReadCloser
+
 	client *Client
 	name   string
 	info   os.FileInfo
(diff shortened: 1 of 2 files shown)


colinmarc:
Hi @flashpixx, thanks for the contribution. What kind of code does this enable? `FileWriter`s should already be usable as `WriteCloser`s, as far as I can tell.

flashpixx:
I'm using your library within an interface which returns an `*io.WriterCloser` and I cannot return your object, because you return a pointer `*FileWriter`. IMHO a `*FilerWriter` pointer and `*io.WriteCloser` pointer are not equal (that are different pointer types). There are two opinions you can return a `io.WriteCloser` object (not a pointer), so the inerhitance will work directly or you need to add my two lines, so the pointer type will be equal. But I hink now that is a better solution that you return an `io.WriteCloser` object.

colinmarc:
io.WriteCloser is an interface, not a struct. Returing a pointer to an interface seems a bit odd to me. Anyway, glad it worked out for you!

PR #219, opened 2020-01-27
public GetServerDefaults

AWoelfel:


flashpixx:
I would like to support this very much

PR #220, opened 2020-02-20
Fix two panic cases on Read()

Xuanwo:
I met some panic while reading from hdfs, here are the screenshots

> sorry I can't provide the plain text as I don't have access to those servers

image
image


colinmarc:
Hi @Xuanwo,

Sorry for the late reply!

Are you using the `FileReader` concurrently? It's not intended to be thread-safe. I don't see how this would be possible otherwise.

I'm going to close this for now, but please let me know if I'm wrong.

PR #232, opened 2020-05-29

### LINKED ISSUE #178: Support list and setting extended attributes

harshavardhana:
The scope of this PR to only support user extended attributes,

Fixes #178


colinmarc:
Hi @harshavardhana,

Thanks for the contribution! As I mentioned in #178, I can't accept the PR without tests attached. I'd also prefer to use `github.com/golang/protobuf` instead of `github.com/gogo/protobuf`, for consistency with the rest of the library.

Issue #221, opened 2020-02-21
Defer remote.Close() may lost some problem.


starsz:
Hello, when I use CopyToRemote function to copy my local file to the remote. I found the file in hdfs is empty, but there is no error log. This problem has troubled me for a long time.

While I read the CopyToRemote function code. The function uses [`defer remote.Close()`](https://github.com/colinmarc/hdfs/blob/4424035185e05136c5431fd61eebd5bf71f05cbc/client.go#L231) may lost some errors in the `Close` function. So I think there shouldn't use `defer`. 

If you agree with me. I can contribute to this project. : )

colinmarc:
Hi @starsz,

I think you're right - I can't see any reason `CopyToRemote` should swallow that error. I'd welcome a PR!

colinmarc:
Fixed in 035b2ab41a0b00a1889a6eff845d9b888cf04856.

NickYadance:
> I found the file in hdfs is empty, but there is no error log. This problem has troubled me for a long time.

Same issue with hdfs file being "empty", after calling `CopyToRemote` api which reports no error.

The content of the file is complete, same as the local file if we `hdfs get` or `hdfs cat` it. The file is actually not empty but being under construction.
```
$ hdfs dfs -checksum /path/to/empty/file
checksum: Fail to get checksum, since file /path/to/empty/file is under construction.
```

The workaround is to recover file lease manually.
```
$ hdfs debug recoverLease -path /path/to/empty/file
recoverLease SUCCEEDED on /path/to/empty/file
```

Unfortunately no stable reproduce method found yet, the guess is that maybe `remote.Close` missed some error conditions.
https://github.com/colinmarc/hdfs/blob/d5784c387f661b5c0044383f0095235fd53c06bd/client.go#L289-L307

PR #234, opened 2020-06-11
Imporve: don't swallow close errors from CopyToRemote and CopyToRemote

starsz:
Hello,  I expose the `Close()` error to function caller, and change the function calling order in `CopyToLocal`, because  I think this may save some network overhead.

ISSUE: https://github.com/colinmarc/hdfs/issues/221


colinmarc:
Merged in a3bc56d. Thanks!

Issue #145, opened 2018-09-11
HDFS data transfer encryption support


mxk1235:
I was attempting to use this library against an hdfs cluster that has `hadoop.rpc.protection` setting in core-site.xml set to `privacy`, as well as  `dfs.encrypt.data.transfer` which is `true` in hdfs-site.xml.   i believe those apply to the protobuf rpc interface.  

the error message i received was the following. 
```no available namenodes: SASL handshake: wrong Token ID. Expected 0504, was 6030  ```

after some debugging i think it occurs here https://github.com/colinmarc/hdfs/blob/master/internal/rpc/kerberos.go#L67 . i suspect the namenode is replying with an encrypted message, while the doKerberosHandshake() expects otherwise.

on first look the library just sets the default value for  `dfs.encrypt.data.transfer`  property as `false`
https://github.com/colinmarc/hdfs/blob/f87e1d64bc48c85b07cab32d23c97788e885b31b/internal/protocol/hadoop_hdfs/hdfs.proto#L402
 and there is no way of creating a client with that property set to true. 
https://github.com/colinmarc/hdfs/blob/f87e1d64bc48c85b07cab32d23c97788e885b31b/client.go#L122

there is a fetchDefaults() function, that's only invoked by file_writer, but not by file_reader (e.g. Stat(), Readdir(), and Read() methods).

can you comment if i'm digging in the right place, and whether the encrypted part of the protocol applies to the read functionality?

here are relevant properties from core-site.xml

```
  <property>
    <name>hadoop.security.authentication</name>
    <value>kerberos</value>
  </property>
  <property>
    <name>hadoop.security.authorization</name>
    <value>true</value>
  </property>
  <property>
    <name>hadoop.rpc.protection</name>
    <value>privacy</value>
  </property>
```

and hdfs-site.xml

```
  <property>
    <name>dfs.encrypt.data.transfer.algorithm</name>
    <value>3des</value>
  </property>
  <property>
    <name>dfs.encrypt.data.transfer.cipher.suites</name>
    <value>AES/CTR/NoPadding</value>
  </property>
  <property>
    <name>dfs.encrypt.data.transfer.cipher.key.bitlength</name>
    <value>256</value>
  </property>
  <property>
    <name>dfs.namenode.acls.enabled</name>
    <value>true</value>
  </property>
```

colinmarc:
Hi there,

Thanks for the report!

[...]

PR #236, opened 2020-07-01
add support for data transfer encryption via rc4 and aes

zeroshade:
addresses #145 

I've only implemented rc4 encryption here as i haven't figured out 3des / des yet, but this at least solves the use case in my own environment which is nice. 

For references for the implementation I used: 
* https://www.ietf.org/rfc/rfc2831.txt
* libgsasl
* libhdfs3

I was able to test this with my own set up using encrypted data transfer and it works! huzzah!

### DIFF
--- a/client.go
+++ b/client.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/colinmarc/hdfs/v2/hadoopconf"
+	"github.com/colinmarc/hdfs/v2/internal/protocol/hadoop_common"
 	hdfs "github.com/colinmarc/hdfs/v2/internal/protocol/hadoop_hdfs"
 	"github.com/colinmarc/hdfs/v2/internal/rpc"
 	krb "gopkg.in/jcmturner/gokrb5.v7/client"
@@ -68,6 +69,13 @@ type ClientOptions struct {
 	// multi-namenode setup (for example: 'nn/_HOST'). It is required if
 	// KerberosClient is provided.
 	KerberosServicePrincipleName string
+	// EncryptDataTransfer specifies whether or not data transfer for datanodes
+	// needs to utilize encryption and thus do a negotiation to get data
+	// this is specified by dfs.encrypt.data.transfer in the config
+	EncryptDataTransfer bool
+	// SecureDataNode specifies whether we're protecting our data transfer
+	// communication via dfs.data.transfer.protection
+	SecureDataNode bool
 }
 
 // ClientOptionsFromConf attempts to load any relevant configuration options
@@ -116,6 +124,26 @@ func ClientOptionsFromConf(conf hadoopconf.HadoopConf) ClientOptions {
 		options.KerberosServicePrincipleName = strings.Split(conf["dfs.namenode.kerberos.principal"], "@")[0]
 	}
 
+	dataTransferProt := strings.Split(strings.ToLower(conf["dfs.data.transfer.protection"]), ",")
+	for _, val := range dataTransferProt {
+		switch val {
+		case "privacy":
+			options.EncryptDataTransfer = true
+			fallthrough
+		case "integrity", "authentication":
+			options.SecureDataNode = true
+		}
+	}
+
+	// dfs.encrypt.data.transfer set to true overrides dfs.data.transfer.protection and
+	// requires both privacy and integrity for communication. If dfs.encrypt.data.transfer is
+	// "true" we explicitly set EncryptDataTransfer and SecureDataNode to true, regardless of
+	// what they already were.
+	if conf["dfs.encrypt.data.transfer"] == "true" {
+		options.EncryptDataTransfer = true
+		options.SecureDataNode = true
+	}
+
 	return options
 }
 
@@ -146,6 +174,20 @@ func NewClient(options ClientOptions) (*Client, error) {
 	}
 
 	return &Client{namenode: namenode, options: options}, nil
+}
+
+func (c *Client) datanodeDialFunc(token *hadoop_common.TokenProto) func(ctx context.Context, network, addr string) (net.Conn, error) {
+	if c.options.EncryptDataTransfer || c.options.SecureDataNode {
+		return (&rpc.DatanodeSaslDialer{
+			Dialer:    c.options.DatanodeDialFunc,
+			Key:       c.namenode.GetEncryptionKeys(),
+			Privacy:   c.options.EncryptDataTransfer,
+			Integrity: c.options.SecureDataNode,
+			Token:     token,
+		}).DialContext
+	}
+
+	return c.options.DatanodeDialFunc
 }
 
 // New returns Client connected to the namenode(s) specified by address, or an
--- a/internal/rpc/block_write_stream.go
+++ b/internal/rpc/block_write_stream.go
@@ -330,29 +330,22 @@ func (s *blockWriteStream) writePacket(p outboundPacket) error {
 		DataLen:           proto.Int32(int32(len(p.data))),
 	}
 
-	header := make([]byte, 6)
+	// Don't ask me why this doesn't include the header proto...
+	totalLength := len(p.data) + len(p.checksums) + 4
+
+	header := make([]byte, 6, 6+totalLength)
 	infoBytes, err := proto.Marshal(headerInfo)
 	if err != nil {
 		return err
 	}
 
-	// Don't ask me why this doesn't include the header proto...
-	totalLength := len(p.data) + len(p.checksums) + 4
 	binary.BigEndian.PutUint32(header, uint32(totalLength))
 	binary.BigEndian.PutUint16(header[4:], uint16(len(infoBytes)))
 	header = append(header, infoBytes...)
+	header = append(header, p.checksums...)
+	header = append(header, p.data...)
 
 	_, err = s.conn.Write(header)
-	if err != nil {
-		return err
-	}
-
-	_, err = s.conn.Write(p.checksums)
-	if err != nil {
-		return err
-	}
-
-	_, err = s.conn.Write(p.data)
 	if err != nil {
 		return err
 	}
--- a/internal/rpc/challenge.go
+++ b/internal/rpc/challenge.go
@@ -32,14 +32,12 @@ type tokenChallenge struct {
 	algorithm string
 }
 
-// parseChallenge returns a tokenChallenge parsed from a challenge response from
-// the namenode.
-func parseChallenge(auth *hadoop.RpcSaslProto_SaslAuth) (*tokenChallenge, error) {
+func parseChallenge(challenge []byte) (*tokenChallenge, error) {
 	tokenChallenge := tokenChallenge{}
 
-	matched := challengeRegexp.FindAllSubmatch(auth.Challenge, -1)
+	matched := challengeRegexp.FindAllSubmatch(challenge, -1)
 	if matched == nil {
-		return nil, fmt.Errorf("invalid token challenge: %s", auth.Challenge)
+		return nil, fmt.Errorf("invalid token challenge: %s", challenge)
 	}
 
 	for _, m := range matched {
@@ -63,4 +61,82 @@ func parseChallenge(auth *hadoop.RpcSaslProto_SaslAuth) (*tokenChallenge, error)
 	}
 
 	return &tokenChallenge, nil
+}
+
+// parseChallengeAuth returns a tokenChallenge parsed from a challenge response from
+// the namenode.
+func parseChallengeAuth(auth *hadoop.RpcSaslProto_SaslAuth) (*tokenChallenge, error) {
+	return parseChallenge(auth.Challenge)
+}
+
+type cipherType uint8
+
+const (
+	cipherUnknown cipherType = 0
+	cipherDES     cipherType = 1 << iota
+	cipher3DES
+	cipherRC4
+	cipherRC440
+	cipherRC456
+	cipherAESCBC
+)
+
+func (c cipherType) String() string {
+	switch c {
+	case cipherDES:
+		return "des"
+	case cipher3DES:
+		return "3des"
+	case cipherRC4:
+		return "rc4"
+	case cipherRC440:
+		return "rc4-40"
+	case cipherRC456:
+		return "rc4-56"
+	case cipherAESCBC:
+		return "aes-cbc"
+	}
+	return ""
+}
+
+func getCipher(s string) cipherType {
+	switch s {
+	case "des":
+		return cipherDES
+	case "3des":
+		return cipher3DES
+	case "rc4":
+		return cipherRC4
+	case "rc4-40":
+		return cipherRC440
+	case "rc4-56":
+		return cipherRC456
+	case "aes-cbc":
+		return cipherAESCBC
+	}
+	return 0
+}
+
+func chooseCipher(cipherOpts []string) cipherType {
+	var avail cipherType
+	for _, c := range cipherOpts {
+		avail |= getCipher(c)
+	}
+
+	if avail&cipherRC4 != 0 {
+		return cipherRC4
+	}
+	// if Has(avail, cipher3DES) {
+	// 	return cipher3DES
+	// }
+	if avail&cipherRC456 != 0 {
+		return cipherRC456
+	}
+	if avail&cipherRC440 != 0 {
+		return cipherRC440
+	}
+	// if Has(avail, cipherDES) {
+	// 	return cipherDES
+	// }
+	return 0
 }
--- a/internal/rpc/digest_md5.go
+++ b/internal/rpc/digest_md5.go
@@ -1,0 +1,426 @@
+package rpc
+
+import (
+	"bytes"
+	"crypto/md5"
+	"encoding/base64"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"hash"
+	"io"
+	"math"
+	"math/rand"
+	"net"
+	"strings"
+
+	"github.com/colinmarc/hdfs/v2/internal/protocol/hadoop_common"
+	hdfs "github.com/colinmarc/hdfs/v2/internal/protocol/hadoop_hdfs"
+)
+
+// EncryptionKey contains the information from the Namenode EncryptionKeys
+// which will be used to establish sasl handshakes
+type EncryptionKey struct {
+	KeyID         int32
+	ExpiryDate    uint64
+	BlockPoolID   string
+	Nonce         []byte
+	EncryptionKey []byte
+	EncryptionAlg string
+}
+
+// our default auth configuration
+var (
+	authMethod    = "TOKEN"
+	authMechanism = "DIGEST-MD5"
+	authServer    = "0"
+	authProtocol  = "hdfs"
+)
+
+// some constants for handling the encoding and decoding of
+// data for sasl communication
+const (
+	saslIntegrityPrefixLength = 4
+	macDataLen                = 4
+	macHMACLen                = 10
+	macMsgTypeLen             = 2
+	macSeqNumLen              = 4
+)
+
+var macMsgType = [2]byte{0x00, 0x01}
+
+// lenEncodeBytes writes the input integer encoded as a 4 byte slice.
+func lenEncodeBytes(seqnum int) (out [4]byte) {
+	out[0] = byte((seqnum >> 24) & 0xFF)
+	out[1] = byte((seqnum >> 16) & 0xFF)
+	out[2] = byte((seqnum >> 8) & 0xFF)
+	out[3] = byte(seqnum & 0xFF)
+	return
+}
+
+type dataTransferStatus *hdfs.DataTransferEncryptorMessageProto_DataTransferEncryptorStatus
+
+// sendSaslMsg is a helper function to encode and write a message for our sasl communication
+func sendSaslMsg(w io.Writer, status dataTransferStatus, payload []byte, message string, secure bool) error {
+	msg := &hdfs.DataTransferEncryptorMessageProto{}
+	msg.Status = status
+	msg.Payload = payload
+	msg.Message = &message
+
+	if secure {
+		// if we want a secure connection, tell the server that we want AES
+		opt := &hdfs.CipherOptionProto{}
+		opt.Suite = hdfs.CipherSuiteProto_AES_CTR_NOPADDING.Enum()
+		msg.CipherOption = append(msg.CipherOption, opt)
+	}
+
+	data, err := makePrefixedMessage(msg)
+	if err != nil {
+		return err
+	}
+
+	_, err = w.Write(data)
+	return err
+}
+
+// digestMD5 holds the information for the token-digestmd5 authentication
+// flow for us to handle the challenges and get the integrity and privacy
+// key pairs for encoding and decoding messages for the server
+type digestMD5 struct {
+	authID   []byte
+	passwd   string
+	hostname string
+	service  string
+
+	token *tokenChallenge
+
+	cnonce string
+	cipher cipherType
+}
+
+// convenience function to calling the a1 function for evaluating challenges
+func (d *digestMD5) a1() string {
+	return a1(string(d.authID), d.token.realm, d.passwd, d.token.nonce, d.cnonce)
+}
+
+// convenience function for calling the a2 function for evaluating challenges
+func (d *digestMD5) a2(first bool) string {
+	return a2(d.service+"/"+d.hostname, d.token.qop, first)
+}
+
+// challengeStep1 implements Using Digest Authentication as a SASAL Mechanism
+// as per RFC2831.
+func (d *digestMD5) challengeStep1(challenge []byte) (string, error) {
+	ncval := fmt.Sprintf("%08x", 1)
+	var err error
+	d.token, err = parseChallenge(challenge)
+	if err != nil {
+		return "", err
+	}
+
+	d.cnonce, err = genCnonce()
+	if err != nil {
+		return "", err
+	}
+
+	d.cipher = chooseCipher(d.token.cipher)
+	rspdigest := compute(d.a1(), d.a2(true), d.token.nonce, d.cnonce, ncval, d.token.qop)
+
+	ret := fmt.Sprintf(`username="%s", realm="%s", nonce="%s", cnonce="%s", nc=%08x, qop=%s, digest-uri="%s/%s", response=%s, charset=utf-8`,
+		d.authID, d.token.realm, d.token.nonce, d.cnonce, 1, d.token.qop, d.service, d.hostname, rspdigest)
+
+	if d.cipher != cipherUnknown {
+		ret += ", cipher=" + d.cipher.String()
+	}
+	return ret, nil
+}
+
+// challengeStep2 implements Using Digest Authentication as a SASAL Mechanism
+// as per RFC2831.
+func (d *digestMD5) challengeStep2(challenge []byte) (string, error) {
+	ncval := fmt.Sprintf("%08x", 1)
+	rspdigest := compute(d.a1(), d.a2(false), d.token.nonce, d.cnonce, ncval, d.token.qop)
+
+	rspauth := strings.Split(string(challenge), "=")
+	if rspauth[0] != "rspauth" {
+		return "", fmt.Errorf("Could not find rspauth in '%s'", string(challenge))
+	}
+	if rspauth[1] != rspdigest {
+		return "", fmt.Errorf("rspauth did not match digest")
+	}
+	return "", nil
+}
+
+// make this easy to mock out for testing by having it defined here where
+// unit tests can override it
+var genCnonce = func() (string, error) {
+	ret := make([]byte, 12)
+	if _, err := rand.Read(ret); err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(ret), nil
+}
+
+// newDigestMD5Conn performs the necessary handshake based on the configuration and then returns
+// a wrapped connection or error, in the case of our auth set to only authentication, then it just
+// returns the connection unmodified after authenticating
+func newDigestMD5Conn(conn net.Conn, token *hadoop_common.TokenProto, key *EncryptionKey, useSecure bool) (net.Conn, error) {
+	auth := &hadoop_common.RpcSaslProto_SaslAuth{}
+	auth.Method = &authMethod
+	auth.Mechanism = &authMechanism
+	auth.ServerId = &authServer
+	auth.Protocol = &authProtocol
+
+	ourToken := &hadoop_common.TokenProto{}
+	ourToken.Kind = token.Kind
+	ourToken.Password = token.Password[:]
+	ourToken.Service = token.Service
+	ourToken.Identifier = token.GetIdentifier()
+
+	// if the server didn't send us a Nonce, then the data isn't encrypted
+	// but we will still attempt to authenticate
+	if useSecure && len(key.Nonce) == 0 {
+		useSecure = false
+	}
+
+	if useSecure {
+		nonceBase64Len := int(math.Ceil(4 * (float64(len(key.Nonce)) / 3)))
+		buf := new(bytes.Buffer)
+		buf.Grow(6 + nonceBase64Len + len(key.BlockPoolID))
+		buf.WriteString(fmt.Sprintf("%d", key.KeyID))
+		buf.WriteString(" ")
+		buf.WriteString(key.BlockPoolID)
+		buf.WriteString(" ")
+		buf.WriteString(base64.StdEncoding.EncodeToString(key.Nonce))
+		ourToken.Identifier = buf.Bytes()
+		ourToken.Password = key.EncryptionKey
+	} else {
+		ourToken.Identifier = make([]byte, base64.StdEncoding.EncodedLen(len(token.GetIdentifier())))
+		base64.StdEncoding.Encode(ourToken.Identifier, token.GetIdentifier())
+	}
+
+	dgst := digestMD5{
+		authID:   ourToken.Identifier,
+		passwd:   base64.StdEncoding.EncodeToString(ourToken.Password),
+		hostname: auth.GetServerId(),
+		service:  auth.GetProtocol(),
+	}
+
+	// start handshake
+	var err error
+	if err = binary.Write(conn, binary.BigEndian, uint32(0xDEADBEEF)); err != nil {
+		return nil, err
+	}
+
+	if err = sendSaslMsg(conn, hdfs.DataTransferEncryptorMessageProto_SUCCESS.Enum(), []byte{}, "", false); err != nil {
+		return nil, err
+	}
+
+	msg := &hdfs.DataTransferEncryptorMessageProto{}
+	if err = readPrefixedMessage(conn, msg); err != nil {
+		return nil, err
+	}
+
+	challengeResponse, err := dgst.challengeStep1(msg.Payload)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = sendSaslMsg(conn, hdfs.DataTransferEncryptorMessageProto_SUCCESS.Enum(), []byte(challengeResponse), "", useSecure); err != nil {
+		return nil, err
+	}
+
+	if err = readPrefixedMessage(conn, msg); err != nil {
+		return nil, err
+	}
+
+	// the result of this challenge evaluation should be an empty string
+	if challengeResponse, err = dgst.challengeStep2(msg.Payload); err != nil || len(challengeResponse) != 0 {
+		return nil, err
+	}
+
+	// we're only authenticating, no RPC protection involved, so no wrapping the connection
+	// just return the connection as is, we're done here!
+	if dgst.token.qop == qopAuthentication {
+		return conn, nil
+	}
+
+	kic, kis := generateIntegrityKeyPair(dgst.a1())
+
+	var d digestMD5Handler
+	if dgst.token.qop == qopPrivacy {
+		if dgst.cipher == cipherUnknown {
+			return nil, fmt.Errorf("Could not find implemented cipher among choices: %v", dgst.token.cipher)
+		}
+		kcc, kcs := generatePrivacyKeyPair(dgst.a1(), dgst.cipher)
+		d, err = newDigestMD5PrivacyConn(conn, kic[:], kis[:], kcc[:], kcs[:])
+		if err != nil {
+			return nil, err
+		}
+	} else if dgst.token.qop == qopIntegrity {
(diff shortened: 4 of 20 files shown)


zeroshade:
Hmm @colinmarc  I haven't been able to reproduce the failure being seen on the Travis run and have no idea what is causing it, any ideas as to what i can do to try to track it down? :)

colinmarc:
Wooh nice! This looks pretty good - I think shifting some abstractions may be in order, though. And we'll need to figure out some testing. Can you add a case to travis.yml for data transfer encryption?

The Append error is unrelated.

zeroshade:
I'll see about adding a case to the travis.yml for testing the data transfer encryption, let me know what shifting around of the abstractions you're thinking of. There were a couple aspects here that i wasn't completely sure about design-wise, but I wanted to keep things as unobtrusive as possible.

My testing was also only against Hadoop 2.X and all the test cases on travis failing with the Append thing and timing out are for CDH6, which is Hadoop 3.X, is there any way we can resolve that Append and timing out issue so that I can ensure that the problem isn't some incompatibility with my change and hadoop 3.x?

colinmarc:
Go ahead and add `t.Skip()` to the append tests that are failing. I have been trying to figure those out for a while and will do so in a different branch..

zeroshade:
W00t! Figured out the cause of the failure and put a fix in. :) I'll play around with getting a good test case added for the case of encrypted data tomorrow

zeroshade:
@colinmarc unit test added for the digestmd5, travis updated with test cases for the data encryption transfer which are now passing and even found a couple bugs in my implementation that i've since fixed.

I see a couple avenues for cleaning up some of the abstractions and interfaces so i'll take a look at that, in the meantime if you have any suggestions let me know.

zeroshade:
Hey @colinmarc, anything else you need from me here to get this merged?

colinmarc:
Hey @zeroshade,

Thanks for the ping. I’ll give this a review in the next few days (if I don’t, please harass me). Excited to have this feature!

Colin

zeroshade:
@colinmarc harass harass harass harass harass :)

consider this me harassing you :)

colinmarc:
Ok, looking through - I think the raw material is really good here but I think it would be better if the functionality was broken up a bit into composable pieces.

Here's how I would organize this: 

 - Rather than having a `DataTransporter` struct, just give `blockReader`/`Writer`/etc a `net.Conn` (which does the wrapping we need), and let them be blissfully unaware. 
 - Avoid wrapping the connection in the default unencrypted case. This is important to avoid the allocation and `copy` in `Decode`/`Encode`, which will probably have a negative impact on performance.
 - While you're at it, reusing a buffer in `Decode`/`Encode` is probably called for.
 - Break `DigestMD5` into `digestMD5PrivacyConn` and `digestMD5IntegrityConn`, or, if those are impossible to untangle, just make the `digestMd5Conn` struct implement `net.Conn` directly. (It's actually not clear to me from [the rfc](https://tools.ietf.org/html/rfc2831#section-2.3) whether those two things can be combined, and what that looks like, but we're not handling that case here anyway.)

Some other general comments:

 - I saw one or two cases where we silently fall back to sending unencrypted data if negotiation fails. That's a really big security hole, and we definitely shouldn't do that.
 - Please avoid exporting names (eg `NewDigestMD5`, `DataTransporter`) that don't need to be exported.
 - Please write comments in full sentences and with the [godoc conventions](https://blog.golang.org/godoc) in mind. Sorry for being a stickler about that, and don't stress about it overmuch - I will make a comment pass before I merge in any case.

Thanks so much for your work on this! Based on the number of people that have been asking for it, it's sorely needed.

colinmarc:
One more performance suggestion: note that [`XORKeyStream`](https://golang.org/pkg/crypto/rc4/#Cipher.XORKeyStream) wants an overlapping dst and src.

zeroshade:
The reason i created the `DataTransporter` struct was because I wanted to consolidate the connection logic rather than duplicating the `key`, `secure` and `token` members in each of `blockReader`/`blockWriter`/etc. Those structs already had a `net.Conn` object, the `DataTransporter` struct just gives them a `connect` function which will handle or wrap the `net.Conn` object in order to let them all be completely unaware because i realized that the code to do the wrapping ended up being identical in all those cases. I'm not really sure how to do what you're asking without ending up with a bunch of duplication amongst those structs.

> Avoid wrapping the connection in the default unencrypted case. This is important to avoid the allocation and copy in Decode/Encode, which will probably have a negative impact on performance.

If you look at transport.go:127 I don't wrap it in the default case, if you're not using `Privacy` or `Token`, then it won't wrap it. In the meantime i'll see about reusing the buffers where I can.

> I saw one or two cases where we silently fall back to sending unencrypted data if negotiation fails. That's a really big security hole, and we definitely shouldn't do that.

That's the situation where the client config is specifying encryption and the server config doesn't match and doesn't respond with the encryption, (which is a common implementation which I pulled from libhdfs3), from a consumer standpoint I'd prefer for it to work than to die when the client config supports encryption but the server has it disabled.

colinmarc:
> I'm not really sure how to do what you're asking without ending up with a bunch of duplication amongst those structs.

I see what you mean. Let's encapsulate that in a `datanodeSaslDialer` closure:
```go
BlockReader{
    DialFunc: (&datanodeSaslDialer {
        token: block.GetBlockToken(),
        ...
    }).DialContext
    ...
}
```

I don't _think_ it's abusing that interface to do negotiation as part of the dial; that seems to be a [pattern](https://godoc.org/golang.org/x/net/proxy#Dialer) elsewhere.

As a side note, I've been writing rust for a bit and it just about broke my brain to remember that you can put the receiver in a closure like that: https://play.golang.org/p/OXQBmmsdLEG

colinmarc:
> That's the situation where the client config is specifying encryption and the server config doesn't match and doesn't respond with the encryption, (which is a common implementation which I pulled from libhdfs3)...

I really think this is a bad idea, regardless of what libhdfs does. If we're talking to a malicious interloper, they can just respond in plaintext and we'll send them the nuclear access codes anyway. On the other hand, the hadoop docs for `data.transfer.encrypted` say:

> Whether or not actual block data that is read/written from/to HDFS should be encrypted on the wire. This only needs to be set on the NN and DNs, clients will deduce this automatically.

Am I missing something, or is that really poorly designed? I'd say for now we require clients to be explicit, and fail if things don't match up. If someone shows up to complain that we're missing a feature for "deducing" server encryption settings, then maybe they can explain to me why it's fine.

zeroshade:
@colinmarc From a consumer standpoint it can make sense, essentially that you don't need to worry about configuring the client, but rather the client will essentially ask the serve whether or not it wants to use encryption and then will do what the server requests. At the same time, I can agree with where you're coming from, it makes sense for a client to be able to say "I want to use an encrypted connection, and i do not want to allow my data across an unencrypted one". I'm fine with requiring clients to be explicit for now and failing if the settings don't match up, I'll make that change.

zeroshade:
So quick question, when you say:

>While you're at it, reusing a buffer in Decode/Encode is probably called for.

Do you mean reusing the input buffer for the output? I'm not sure how i feel about that one as far as overwriting the input buffer with the modified data, I think I prefer doing the copy.  Or was there a different reusing you were referring to?

colinmarc:
I meant that we shouldn't be doing _any_ allocation inside `net.Write` or `net.Read`, since that's a super duper hot path. The best way to reuse an allocation is with [`bytes.Buffer`](https://golang.org/pkg/bytes/#Buffer). So instead of:

```go
b := make([]byte, n)
conn.Read(b)
```

It's much better to have a `bytes.Buffer` saved on the struct, and do something like (on read, for example):

```go
d.buf.Reset()
io.Copy(d.buf, conn) // This will use ReadFrom under the hood
b := d.buf.Bytes()
```

zeroshade:
Ah, gotcha. that makes sense.

zeroshade:
Hmm, @colinmarc any reason you can think of to not just do the check and wrap of the dialer inside of `NewClient`? Looking at stuff, i think that If I wrap the provided `DatanodeDialFunc` in the `ClientOptions` during `NewClient` then basically everything gets it for free without having to care :)

colinmarc:
> any reason you can think of to not just do the check and wrap of the dialer inside of NewClient?

Sounds elegant to me!

colinmarc:
If you rebase, you can remove the `t.Skip` for those append tests.

zeroshade:
@colinmarc I still need to go through and put more comments in and such, but i'd like your opinion on the refactor / redesign based on our conversations. Let me know what you think. thanks!

zeroshade:
Hmmm don't know why that `writeLeaseRenewal` function is failing the test now, I'll have to look at this is the morning

zeroshade:
looked at master, that test is currently skipped. so the failure is not my fault :) i just got overzealous when you said i could rebase and remove the skips haha. I've put the skip back and hopefully everything should succeed now.

zeroshade:
@colinmarc Finally! :) Passing all tests, and redesigned based on your comments. Consider this me harassing you to take a look and let me know what you think now :)

client.go
```diff
@@ -68,6 +69,13 @@ type ClientOptions struct {
 	// multi-namenode setup (for example: 'nn/_HOST'). It is required if
 	// KerberosClient is provided.
 	KerberosServicePrincipleName string
+	// EncryptedDataNode specifies whether or not data transfer for datanodes
+	// needs to utilize encryption and thus do a negotiation to get data
+	// this is specified by dfs.encrypt.data.transfer in the config
+	EncryptedDataNode bool
+	// SecureDataNode specifies whether we're using block access tokens to
+	// communicate with datanodes, specified by dfs.block.access.token.enable in the config
+	SecureDataNode bool
```
colinmarc, line 78: I thought this was something else, actually? We've always sent the block token, and I think this is just a server-side option for verifying them (with or without kerberos). In any case it should be `EnableBlockAccessToken`

client.go
```diff
@@ -116,6 +124,22 @@ func ClientOptionsFromConf(conf hadoopconf.HadoopConf) ClientOptions {
 		options.KerberosServicePrincipleName = strings.Split(conf["dfs.namenode.kerberos.principal"], "@")[0]
 	}
 
+	dataTransferProt := strings.ToLower(conf["dfs.data.transfer.protection"])
```
colinmarc, line 127: Technically, this is a comma-separated list, so you'll need to split it.

client.go
```diff
@@ -116,6 +124,22 @@ func ClientOptionsFromConf(conf hadoopconf.HadoopConf) ClientOptions {
...
+		options.SecureDataNode = true
+	}
+
+	// dfs.encrypt.data.transfer set to true overrides dfs.data.transfer.protection and
+	// requires both privacy and integrity for communication
+	if conf["dfs.encrypt.data.transfer"] == "true" {
+		options.EncryptedDataNode = true
+		options.SecureDataNode = true
+	}
+
```
colinmarc, line 142: The comment for this method needs to be updated (to say that we munge these fields)

client.go
```diff
@@ -148,6 +172,16 @@ func NewClient(options ClientOptions) (*Client, error) {
...
 
+func (c *Client) datanodeDialFunc(token *hadoop_common.TokenProto) func(ctx context.Context, network, addr string) (net.Conn, error) {
+	return (&rpc.DatanodeSaslDialer{
+		Dialer:    c.options.DatanodeDialFunc,
+		KeyFunc:   c.namenode.GetEncryptionKeys,
+		Privacy:   c.options.EncryptedDataNode,
+		Integrity: c.options.SecureDataNode,
+		Token:     token,
+	}).DialContext
+}
```
colinmarc, line 183: Can we avoid wrapping it at all if those options are false?

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,473 @@
...
+	net.Conn
+	decode(input []byte) ([]byte, error)
+}
+
+// so we don't have to duplicate this for all three types of connection
+// wrappers (md5 integrity, md5 privacy, and aes), have all of them delegate
+// the functions we aren't overriding to the underlying connection object
+// but we don't override the Read and Write here forcing each of the types
+// to have to implement their own in order to meet the net.Conn interface
+type baseConnWrapper struct {
```
colinmarc, line 298: Let's just duplicate this. Also, why do we need to track the deadline?

internal/rpc/block_read_stream.go
```diff
@@ -101,8 +101,8 @@ func (s *blockReadStream) Read(b []byte) (int, error) {
 	for i := 0; i < chunksToRead; i++ {
 		chunkOff := i * s.chunkSize
 		chunkEnd := chunkOff + s.chunkSize
-		if chunkEnd >= len(b) {
-			chunkEnd = len(b)
+		if chunkEnd >= n {
```
colinmarc, line 104: This looks right, but what led you to change it? I should maybe push it upstream separately so we can track if it causes any issues.

internal/rpc/block_reader.go
```diff
@@ -28,8 +28,7 @@ type BlockReader struct {
 	// UseDatanodeHostname specifies whether the datanodes should be connected to
 	// via their hostnames (if true) or IP addresses (if false).
 	UseDatanodeHostname bool
-	// DialFunc is used to connect to the datanodes. If nil, then
-	// (&net.Dialer{}).DialContext is used.
+	// DialFunc is used to connect to the datanodes. If nil, then (&net.Dialer{}).DialContext is used
```
colinmarc, line 31: Please keep the newline (and period)

internal/rpc/block_write_stream.go
```diff
@@ -341,22 +341,14 @@ func (s *blockWriteStream) writePacket(p outboundPacket) error {
 	binary.BigEndian.PutUint32(header, uint32(totalLength))
 	binary.BigEndian.PutUint16(header[4:], uint16(len(infoBytes)))
 	header = append(header, infoBytes...)
+	header = append(header, p.checksums...)
+	header = append(header, p.data...)
```
colinmarc, line 345: Can you give `header` some extra capacity up on line 333?

internal/rpc/block_writer.go
```diff
@@ -34,8 +34,7 @@ type BlockWriter struct {
 	// UseDatanodeHostname indicates whether the datanodes will be connected to
 	// via hostname (if true) or IP address (if false).
 	UseDatanodeHostname bool
-	// DialFunc is used to connect to the datanodes. If nil, then
-	// (&net.Dialer{}).DialContext is used.
+	// DialFunc is used to connect to the datanodes. If nil, then (&net.Dialer{}).DialContext is used
```
colinmarc, line 37: Did your code formatter do this?

internal/rpc/challenge.go
```diff
@@ -64,3 +62,86 @@ func parseChallenge(auth *hadoop.RpcSaslProto_SaslAuth) (*tokenChallenge, error)
...
+	cipherUnknown cipherType = 0
+	cipherDES     cipherType = 1 << iota
+	cipher3DES
+	cipherRC4
+	cipherRC440
+	cipherRC456
+	cipherAESCBC
+)
+
+func set(cur, c cipherType) cipherType    { return cur | c }
```
colinmarc, line 84: Just inline these, it's clearer.

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,473 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	d.cnonce, err = genCnonce()
+	if err != nil {
+		return "", err
+	}
+
+	d.cipher = chooseCipher(d.token.cipher)
```
colinmarc, line 128: I think we need to error here if we don't support the given cipher.

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,473 @@
...
+// lenEncodeBytes does what writeLenEncode does, but manipulates the first
+// four bytes of a slice rather than writing to a buffer
+func lenEncodeBytes(seqnum int, out []byte) {
+	out[0] = byte((seqnum >> 24) & 0xFF)
+	out[1] = byte((seqnum >> 16) & 0xFF)
+	out[2] = byte((seqnum >> 8) & 0xFF)
+	out[3] = byte(seqnum & 0xFF)
+}
+
+// just an alias for shorthand since it's such a long name
```
colinmarc, line 62: full sentences please (with punctuation!)

colinmarc: (in this case, just baleet the comment altogether)

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,473 @@
...
+	d.readDeadline = t
+	return d.conn.SetReadDeadline(t)
+}
+
+func (d *baseConnWrapper) SetWriteDeadline(t time.Time) error {
+	d.writeDeadline = t
+	return d.conn.SetWriteDeadline(t)
+}
+
+// Encode will encode a message from the client to send to the server
```
colinmarc, line 332: stray comment?

internal/rpc/digest_md5_integrity.go
```diff
@@ -0,0 +1,121 @@
...
+func newDigestMD5IntegrityConn(conn net.Conn, kic, kis []byte) digestMD5Handler {
+	return &digestMD5IntegrityConn{
+		baseConnWrapper: baseConnWrapper{conn: conn},
+		kic:             kic,
+		kis:             kis,
+	}
+}
+
+func (d *digestMD5IntegrityConn) Write(b []byte) (n int, err error) {
+	if !d.writeDeadline.IsZero() && d.writeDeadline.Before(time.Now()) {
```
colinmarc, line 33: I don't think this is necessary? The underlying call to `Write` should take care of that for us.

internal/rpc/digest_md5_integrity.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	inputLen := len(b)
+	var seqBuf [4]byte
+	lenEncodeBytes(d.sendSeqNum, seqBuf[:])
+
+	outputLen := macDataLen + inputLen + macHMACLen + macMsgTypeLen + macSeqNumLen
+
+	n = 4
+	err = binary.Write(d.conn, binary.BigEndian, int32(outputLen-macDataLen))
```
colinmarc, line 44: Too many calls to `conn.Write` here - we should try to just do one. Better to write to a buffer and then `io.Copy` it out.

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,473 @@
...
+}
+
+// If message integrity is negotiated, there will be a MAC block for each message appended
+// to the message. The MAC block is 16 bytes, the first 10 bytes of the HMAC-MD5 of the message
+// plus a 2 byte message type number and a 4 byte sequence number. This function provides the
+// HMAC as defined in the RFC as HMAC(ki, {seqnum, msg})[0..9]
+func getHMAC(key, seq, msg []byte) []byte {
+	data := append(seq, msg...)
+
+	mac := hmac.New(md5.New, key)
```
colinmarc, line 428: This is unfortunately hiding an allocation: https://golang.org/src/crypto/hmac/hmac.go?s=1932:1982#L60

We can reuse one and save it on the struct.

colinmarc, line 428: This is unfortunately hiding an allocation: https://golang.org/src/crypto/hmac/hmac.go?s=1932:1982#L60

We can reuse one and save it on the struct.

internal/rpc/digest_md5_integrity.go
colinmarc: (It's not really about hyper-optimizing, but we can afford to be hygienic here)

internal/rpc/digest_md5_integrity.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (d *digestMD5IntegrityConn) Write(b []byte) (n int, err error) {
+	if !d.writeDeadline.IsZero() && d.writeDeadline.Before(time.Now()) {
+		return 0, syscall.ETIMEDOUT
+	}
+
+	inputLen := len(b)
+	var seqBuf [4]byte
+	lenEncodeBytes(d.sendSeqNum, seqBuf[:])
```
colinmarc, line 39: I think you can just have this return a [4]byte (but I haven't tried it). Doing so makes me more confident the value won't escape to the heap.

internal/rpc/digest_md5_integrity.go
```diff
@@ -0,0 +1,121 @@
...
+		return 0, syscall.ETIMEDOUT
+	}
+
+	n, err := d.buf.Read(b)
+	if len(b) == n || (err != nil && err != io.EOF) {
+		return n, err
+	}
+
+	var sz int32
+	binary.Read(d.conn, binary.BigEndian, &sz)
```
colinmarc, line 82: You need to check the error here.

internal/rpc/digest_md5_integrity.go
```diff
@@ -0,0 +1,121 @@
...
+
+	d.buf.Reset()
+	d.buf.Grow(int(sz))
+	io.CopyN(&d.buf, d.conn, int64(sz))
```
colinmarc, line 86: Make sure you check the error here, too.

internal/rpc/digest_md5_integrity.go
```diff
@@ -0,0 +1,121 @@
...
+
+	dataLen := inputLen - macHMACLen - macMsgTypeLen - macSeqNumLen
+	hmac := getHMAC(d.kis, seqBuf[:], input[:dataLen])
+
+	seqNumStart := inputLen - macSeqNumLen
+	msgTypeStart := seqNumStart - macMsgTypeLen
+	origHashStart := msgTypeStart - macHMACLen
+
+	if !bytes.Equal(hmac, input[origHashStart:origHashStart+macHMACLen]) ||
+		!bytes.Equal(macMsgType[:], input[msgTypeStart:msgTypeStart+macMsgTypeLen]) ||
```
colinmarc, line 114: Is the `[:]` necessary?

internal/rpc/digest_md5_privacy.go
```diff
@@ -0,0 +1,138 @@
...
+
+	var sz int32
+	err = binary.Read(d.conn, binary.BigEndian, &sz)
+	if err != nil {
+		return 0, err
+	}
+
+	d.readBuf.Reset()
+	d.readBuf.Grow(int(sz))
+	io.CopyN(&d.readBuf, d.conn, int64(sz))
```
colinmarc, line 73: Check the error here.

internal/rpc/digest_md5_integrity.go
colinmarc: This is also probably inefficient (it'll do a tiny unaligned read) but using a buffered read might be more trouble than it's worth (it'll muddle deadlines).

internal/rpc/digest_md5_privacy.go
```diff
@@ -0,0 +1,138 @@
...
+
+	var seqBuf [4]byte
+	lenEncodeBytes(d.readSeqNum, seqBuf[:])
+
+	hmac := getHMAC(d.kis, seqBuf[:], input[:encryptedLen])
+	msgType := input[msgTypeStart : msgTypeStart+macMsgTypeLen]
+	seqNum := input[seqNumStart : seqNumStart+macSeqNumLen]
+
+	if !bytes.Equal(hmac, origHash) || !bytes.Equal(macMsgType[:], msgType) || !bytes.Equal(seqNum, seqBuf[:]) {
+		return nil, fmt.Errorf("HMAC Integrity Check Failed")
```
colinmarc, line 103: `errors.New`

internal/rpc/namenode.go
```diff
@@ -223,6 +223,24 @@ func (c *NamenodeConnection) Execute(method string, req proto.Message, resp prot
 	return nil
 }
 
+// GetEncryptionKeys will use the `getDataEncryptionKey` operation on the
+// namenode in order to fetch the current data encryption keys
+func (c *NamenodeConnection) GetEncryptionKeys() *EncryptionKey {
```
colinmarc, line 228: I'm not sure about exporting this. Does it have utility to users?

internal/rpc/sasl_dialer.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/colinmarc/hdfs/v2/internal/protocol/hadoop_common"
+)
+
+// DatanodeSaslDialer is a dialer that will use the Privacy and Integrity
+// data members in order to decide whether or not it needs to wrap the connection
+// to the datanode for proper handling of encryption / integrity protection
+// after it's been constructed it can be used as one would use net.Dialer
+type DatanodeSaslDialer struct {
+	Dialer    func(ctx context.Context, network, addr string) (net.Conn, error)
+	KeyFunc   func() *EncryptionKey
```
colinmarc, line 16: This is a bit dicey. I'd prefer we call this function once and then reuse the result.

client.go
zeroshade: Turns out you're right, this should be changed to reference `dfs.data.transfer.protection` not `dfs.block.access.token.enable`

client.go
zeroshade: If you look at the definition for `DialContext` in the `DatanodeSaslDialer` it indeed will just call `Dial` and won't wrap the connection if those values are false. If you'd prefer though I can have this just return the DatanodeDialFunc if those values are false rather than using the sasl dialer

internal/rpc/digest_md5.go
zeroshade: Because of the tests for the deadline. Since we buffer the reads to make sure that we receive a whole message to decode properly and then store the remaining bytes in the buffer, it doesn't actually go back to the connection until it runs out of data, therefore it won't actually return the timeout unless it has to go back to the connection which the current tests we have don't do. So i had to add this to have it pass the tests. I think if we increase the size of the data being used for the deadline tests it might be able to pass it instead of us tracking the deadline ourselves.

internal/rpc/block_read_stream.go
zeroshade: it was causing a test to fail because the `b` that was used was sized to 512 but the test is the `foo.txt` which only has 4 bytes. As a result the checksum would fail because chunkEnd was the full chunksize rather than only the 4 bytes for the data.

internal/rpc/block_reader.go
zeroshade: funny story, my gocode linter actually says it *doesn't* want the punctuation at the end lol. I'll put it back though.

internal/rpc/digest_md5_integrity.go
zeroshade: I originally had that but when i was using a buffer that I reused, i ended up with a data race here in the test that tests `block_write_stream`. By having it write to the connection directly we avoid the race condition because there's no intermediate buffer. I really didn't want to try tracing down the race condition that was happening when i reused a write buffer :(

internal/rpc/digest_md5_integrity.go
zeroshade: `bytes.Equal` needs a slice, but we static initialized `macMsgType` as a `[2]byte`, so you need the `[:]` for it to be passed to `bytes.Equal`.

internal/rpc/namenode.go
zeroshade: right now, i kept it exported because it's `Client` which is using it to pass to the `DatanodeSaslDialer` as the `KeyFunc` so that the dialer doesn't need to have a reference to the namenode connection directly, just a function which gives it the key. So since it's being used in the `client.go` file, it needs to be exported.

internal/rpc/sasl_dialer.go
zeroshade: fair, I can call it in `Client` when we create this and store the result rather than passing the function itself.

colinmarc: hm, maybe just pass the two options into `NamenodeConnection` and then put the `wrapDatanodeDialer` func there?

colinmarc: whoops, meant that in response to the above comment

internal/rpc/digest_md5_integrity.go
zeroshade: I'm already basically doing a buffered read and muddling the deadlines, which is why we're tracking them separately on here, since you can't ask the `net.Conn` what it's current deadlines are

internal/rpc/namenode.go
zeroshade: > hm, maybe just pass the two options into NamenodeConnection and then put the wrapDatanodeDialer func there?

but then i'd have to export wrapDatanodeDialer since NamenodeConnection is in `internal/rpc` and this needs to be called from `file_reader` / `file_writer`. Either way something would need to get exposed since the block reader / writer don't have a reference to the namenode, which they shouldn't need anyways.

zeroshade: I figured it made more sense to export this utility function which is just a function call on the Namenode than it would have been to export a wrapping function for the datanode dialer.

internal/rpc/digest_md5_integrity.go
colinmarc: Hm, how could there be a race? The buffer is only used for the duration of this call.

internal/rpc/digest_md5.go
colinmarc: That's a good point. It sounds like we should track it, even if we can trick the test. I think that means you only need to track it for `Read`, though, not for `Write`.

internal/rpc/block_read_stream.go
colinmarc: Which test? I'm going to see if I can add a failing one to master.

internal/rpc/digest_md5_integrity.go
zeroshade: I had put the buffer onto the struct itself like you suggested because you didn't want the allocations in here, so the buffer would be shared between write calls (look at the write implementation in the privacy connection for an example)

internal/rpc/block_read_stream.go
zeroshade: I'm pretty sure it was the client_test.go `TestReadFile` i'd have to double check though.

client.go
zeroshade: comment updated. let me know if this is good.

client.go
zeroshade: Let me know if the new comments / updated name/comment is good.

internal/rpc/digest_md5.go
zeroshade: I've updated to not track the writedeadline and only track the readdeadline.

internal/rpc/block_write_stream.go
zeroshade: done!

internal/rpc/digest_md5_integrity.go
zeroshade: Hmm, I think it might have been me sharing the buffer between the read and writes rather than having a readbuffer and writebuffer. Looks like i got it all squared away now using a buffer here to write to and then using `buffer.WriteTo` to actually write it all to the connection. Looks good to me. let me know what you think.

internal/rpc/digest_md5_integrity.go
zeroshade: done.

zeroshade:
@colinmarc  holy crap all the tests pass still and i've addressed the feedback. W00t w00t :) Just waiting for a few more comments from you on some of the things to let me know if they are good.

internal/rpc/block_read_stream.go
colinmarc: Huh, I can't figure out how to reproduce this issue on master. Do you have any idea what it is the wrapped conn is doing that triggers it?

I just tried this:

```
type randomReadConn struct {
	net.Conn
}

func (r *randomReadConn) Read(b []byte) (int, error) {
	return r.Conn.Read(b[0:rand.Intn(len(b)+1)])
}

func TestFileBigReadWeirdSizesMisalignment(t *testing.T) {
	client := getClient(t)
	dial := client.options.DatanodeDialFunc
	if dial == nil {
		dial = (&net.Dialer{}).DialContext
	}

	client.options.DatanodeDialFunc = func(ctx context.Context, network, address string) (net.Conn, error) {
		conn, err := dial(ctx, network, address)
		if err != nil {
			return nil, err
		}

		return &randomReadConn{conn}, nil
	}

	file, err := client.Open("/_test/mobydick.txt")
	require.NoError(t, err)

	hash := crc32.NewIEEE()
	n, err := io.Copy(hash, file)
	assert.NoError(t, err)
	assert.EqualValues(t, n, 1257276)
	assert.EqualValues(t, 0x199d1ae6, hash.Sum32())
}
```

zeroshade:
@colinmarc well, i'm stumped, something about how i've changed things now is changing the way the allocations happen for the size of that []byte slice, and now it's not exhibiting that issue anymore -_- oy vey. i'm so confused, but at least it all works now haha

colinmarc:
Can you rebase in 15f6da0033bd0247df209c306cac9ab6747ca175 (and squash) to make sure those tests pass too?

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,426 @@
...
+
+	ourToken := &hadoop_common.TokenProto{}
+	ourToken.Kind = token.Kind
+	ourToken.Password = token.Password[:]
+	ourToken.Service = token.Service
+	ourToken.Identifier = token.GetIdentifier()
+
+	// if the server didn't send us a Nonce, then the data isn't encrypted
+	// but we will still attempt to authenticate
+	if useSecure && len(key.Nonce) == 0 {
```
colinmarc, line 182: In which case does the server not send a nonce?

colinmarc:
Just found this in the RFC:

```
Digest authentication is vulnerable to "man in the middle" (MITM)
   attacks. Clearly, a MITM would present all the problems of
   eavesdropping. But it also offers some additional opportunities to
   the attacker.

   A possible man-in-the-middle attack would be to substitute a weaker
   qop scheme for the one(s) sent by the server; the server will not be
   able to detect this attack. For this reason, the client should always
   use the strongest scheme that it understands from the choices
   offered, and should never choose a scheme that does not meet its
   minimum requirements.
```

Sounds like we can just use the server's QOP, and hope we're not mitm'd. I'm munging some comments now, so I'll make the change.

internal/rpc/digest_md5.go
```diff
@@ -0,0 +1,426 @@
...
+		kcc, kcs := generatePrivacyKeyPair(dgst.a1(), dgst.cipher)
+		d, err = newDigestMD5PrivacyConn(conn, kic[:], kis[:], kcc[:], kcs[:])
+		if err != nil {
+			return nil, err
+		}
+	} else if dgst.token.qop == qopIntegrity {
+		d = newDigestMD5IntegrityConn(conn, kic[:], kis[:])
+	}
+
+	if len(msg.GetCipherOption()) > 0 {
```
colinmarc, line 264: Shouldn't this always be true if the qop is `privacy`? Put another way, is `digestMD5PrivacyConn` ever actually used (beyond the call to `decode` below)?

internal/rpc/digest_md5.go
zeroshade: if the client has the data transfer protection set to privacy, but the server is configured for auth-integrity or authentication but not auth-privacy, then there is no key to send and as such the nonce will be empty.  In that case we fall back to what the server's qop is.

internal/rpc/digest_md5.go
zeroshade: if for some reason the server doesn't support AES then this would be empty and it would only use the rc4 that was negotiated during the digest authentication.

zeroshade:
@colinmarc Rebased and Squashed

zeroshade:
@colinmarc 

according to travis this completed successfully, i don't know why it still is listed as in progress here :(

colinmarc:
When I extend the matrix to test different `rpc.protection` and `data.transfer.protection` values, I get test failures: https://travis-ci.org/github/colinmarc/hdfs/jobs/709051530

colinmarc:
Merged :) thanks so much for your work on this!!!

colinmarc:
Just a note I've been trying to track down a flakey test failure in privacy mode - in case you have any insight: https://travis-ci.org/github/colinmarc/hdfs/jobs/709475318

This is the error from the datanode logs:

```
java.lang.NumberFormatException: For input string: "3817551356"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:583)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer.getEncryptionKeyFromUserName(SaslDataTransferServer.java:266)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer.access$000(SaslDataTransferServer.java:73)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer$1.apply(SaslDataTransferServer.java:177)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer$SaslServerCallbackHandler.handle(SaslDataTransferServer.java:241)
	at com.sun.security.sasl.digest.DigestMD5Server.validateClientResponse(DigestMD5Server.java:589)
	at com.sun.security.sasl.digest.DigestMD5Server.evaluateResponse(DigestMD5Server.java:244)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslParticipant.evaluateChallengeOrResponse(SaslParticipant.java:115)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer.doSaslHandshake(SaslDataTransferServer.java:376)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer.getEncryptedStreams(SaslDataTransferServer.java:180)
	at org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferServer.receive(SaslDataTransferServer.java:112)
	at org.apache.hadoop.hdfs.server.datanode.DataXceiver.run(DataXceiver.java:234)
	at java.lang.Thread.run(Thread.java:748)
```

I haven't started looking through the traceback, but I will tomorrow.

zeroshade:
@colinmarc yeah, you can't use the `hdfs.DataEncryptionKeyProto` directly unfortunately. Turns out that even though the protobuf has the key id has a `uint32` you need to treat it as an `int32` during the digest handshake generating the username and token. The key id can be negative -_- that's why I had a separate struct for it

colinmarc:
Aha! I missed that detail when futzing with your code. Sorry about that.

zeroshade:
@colinmarc any chance of seeing a release with this? I'm having to use mod replace to point to the master branch right now since there's no release with the changes yet

PR #238, opened 2020-07-16

### LINKED ISSUE #178: Support list and setting extended attributes

harshavardhana:
The scope of this PR to only support user extended attributes,

Fixes #178


colinmarc:
Fixed in 7dba08248002d73cb14dfa7b848f6b08471ef03d.

Issue #237, opened 2020-07-07
namenode channel should be closed for once


TWL007:
https://github.com/colinmarc/hdfs/blob/cd715bff28aefb46a4e1e24c03c49f9a874818eb/internal/rpc/namenode.go#L302-L315

We shoud make sure channel will be closed for once to avoid panic:
`panic: close of closed channel [recovered]
	panic: close of closed channel `

colinmarc:
See my comment on #241.

PR #241, opened 2020-08-15
Fix close of closed channel problem

TWL007:
The HDFS connection could be closed more than one time. Once we close the closed channel, we will get go panic: `panic: close of closed channel `.

This PR is trying to add sync.Once to make close func will be only called one time to avoid this problem. 

Related issue: #237

### DIFF
--- a/internal/rpc/namenode.go
+++ b/internal/rpc/namenode.go
@@ -48,8 +48,9 @@ type NamenodeConnection struct {
 	hostList  []*namenodeHost
 	transport transport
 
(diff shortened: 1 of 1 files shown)


TWL007:
@colinmarc Hi Colin, could you help me review this PR? Thanks~

colinmarc:
Hi @TWL007,

Sorry for the long-delayed response, but Client objects are not intended to be thread-safe. Calling code should ensure it has one per goroutine.

PR #246, opened 2020-09-23
implementing `ls -d` and `ls -R`

sknopp:
I've added two new features to ls: `-d` and `-R`

Their output behaviour is the same as that of the Unix tool.

### DIFF
--- a/cmd/hdfs/ls.go
+++ b/cmd/hdfs/ls.go
@@ -13,7 +13,9 @@ import (
 	"github.com/colinmarc/hdfs/v2"
 )
 
-func ls(paths []string, long, all, humanReadable bool) {
+var recursionLevel = 0
+
+func ls(paths []string, long, all, humanReadable bool, recursive bool, selfOnly bool) {
 	paths, client, err := getClientAndExpandedPaths(paths)
 	if err != nil {
 		fatal(err)
@@ -41,7 +43,7 @@ func ls(paths []string, long, all, humanReadable bool) {
 	}
 
 	if len(files) == 0 && len(dirs) == 1 {
-		printDir(client, dirs[0], long, all, humanReadable)
+		printDir(client, dirs[0], long, all, humanReadable, recursive, selfOnly)
 	} else {
 		if long {
 			tw := lsTabWriter()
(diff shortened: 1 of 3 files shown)


cmd/hdfs/ls.go
```diff
@@ -13,7 +13,9 @@ import (
 	"github.com/colinmarc/hdfs/v2"
 )
 
-func ls(paths []string, long, all, humanReadable bool) {
+var recursionLevel = 0
```
colinmarc, line 16: Can you reimplement this without a global variable?

cmd/hdfs/ls.go
```diff
@@ -13,7 +13,9 @@ import (
...
+
+func ls(paths []string, long, all, humanReadable bool, recursive bool, selfOnly bool) {
```
colinmarc, line 18: This can be `long, all, humanReadable, recursive, selfOnly bool`

cmd/hdfs/main.go
```diff
@@ -43,6 +43,8 @@ Valid commands:
 	lsl    = lsOpts.Bool('l')
 	lsa    = lsOpts.Bool('a')
 	lsh    = lsOpts.Bool('h')
+	lsd    = lsOpts.Bool('d')
+	lsR    = lsOpts.Bool('R')
```
colinmarc, line 47: Add these to the help text, right above.

cmd/hdfs/ls.go
```diff
@@ -56,18 +58,31 @@ func ls(paths []string, long, all, humanReadable bool) {
 			}
 		}
 
-		for i, dir := range dirs {
-			if i > 0 || len(files) > 0 {
-				fmt.Println()
+		if selfOnly {
```
colinmarc, line 61: Rather than switching here, I think it makes sense to just skip the step above where the dirs are separated out.

colinmarc: That way `printFiles` and `printDirs` don't need to care.

cmd/hdfs/ls.go
```diff
@@ -100,30 +129,67 @@ func printDir(client *hdfs.Client, dir string, long, all, humanReadable bool) {
...
+			if recursionLevel > 0 {
+				fmt.Println()
+			}
+			fmt.Printf("%s:\n", dir)
+			fmt.Println("total", fileCount)
+		}
+		recursionLevel++
 	}
 
+	printFiles(client, tw, dir, files, long, all, recursive, humanReadable, selfOnly)
```
colinmarc, line 159: Instead of recursing here, let's move that functionality higher up and use `Walk`. You can follow the example for `chmod`: https://github.com/colinmarc/hdfs/blob/master/cmd/hdfs/chmod.go#L35-L51

colinmarc:
Hi @sknopp, thank you so much for the contribution! Sorry for the long delay responding. I left some comments.

PR #251, opened 2020-10-28
Export method to get an unique name of the client

vasayxtx:
Hi!
When the application uses several clients, it is very convenient to use the unique names of each of them in logs to make it easier to deal with problems.


colinmarc:
Works for me! Thanks for the contribution.

PR #253, opened 2020-11-15
Process AlreadyBeingCreatedException HDFS error as os.ErrExist on Create

vasayxtx:
Seems like there was a typo in HDFS exception name: the class name started with a small letter instead of a capital one (`org.apache.hadoop.fs.fileAlreadyExistsException` -> `org.apache.hadoop.fs.FileAlreadyExistsException`).

Also, if we try to create a file that already is being created, but is not closed yet, `org.apache.hadoop.hdfs.protocol.AlreadyBeingCreatedException` will occur (https://github.com/apache/hadoop/blob/a55d6bba71c81c1c4e9d8cd11f55c78f10a548b0/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/protocol/AlreadyBeingCreatedException.java#L32). Propose to convert this exception to `os.ErrExist` error too.

### DIFF
--- a/file_writer_test.go
+++ b/file_writer_test.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


vasayxtx:
@colinmarc, could you please review the PR?

file_writer_test.go
```diff
@@ -270,6 +265,39 @@ func TestCreateEmptyFile(t *testing.T) {
...
 
+func TestCreateFileExists(t *testing.T) {
+	const filePath = "/_test/create_file_exists"
+
+	client := getClient(t)
+	createFile := func() (*FileWriter, error) {
+		return client.CreateFile(filePath, 1, 1048576, 0755)
+	}
+
+	t.Run("prev file closed", func(t *testing.T) {
```
colinmarc, line 276: Do you mind writing this to stylistically match the other tests? Sorry for being picky.

file_writer_test.go
```diff
@@ -14,26 +14,21 @@ import (
...
-				t.Log("Ignoring AlreadyBeingCreatedException from append")
-
-				if retries < 3 {
-					retries += 1
-					continue
-				} else {
-					t.Skip("skipping Append test because of repeated AlreadyBeingCreatedException")
-					return fw, nil
-				}
+		if os.IsExist(err) {
```
colinmarc, line 24: I think this should stay the way it is (and only ignore the ABC error, rather than any IsExist).

colinmarc:
@vasayxtx, sorry I'm a month late on this! I left two nitpicky comments, but otherwise looks good.

colinmarc:
I fixed the classname issue in b2bcdb85539bbe670509010e11ae8bd32b2db26f.

file_writer_test.go
vasayxtx: Reworked, sub-tests were rewritten as separate tests.

file_writer_test.go
vasayxtx: `AlreadyBeingCreatedException` was interpreted as `os.ErrExist` for all HDFS requests, so it was impossible to ignore only this exception here. Actually, receiving `os.ErrExist` error on `append` operation is a bit strange, so I decided to change implementation and interpret `AlreadyBeingCreatedException` only on `create`.

vasayxtx:
> I fixed the classname issue in [b2bcdb8](https://github.com/colinmarc/hdfs/commit/b2bcdb85539bbe670509010e11ae8bd32b2db26f).

ok, thanks.

colinmarc:
Merged in f2f512d. Thank you!!

PR #254, opened 2020-11-22
Crc32 sums

urykhy:
default hadoop checksum is not very nice to use.
add option to use crc32.

ps: first commit (protobuf generation) is a fix for 'protoc-gen-go: error:inconsistent package import paths:'.


colinmarc:
Hi @urykhy,

Thanks for the suggestion. While I agree the hadoop checksum scheme isn't very nice, I think we're stuck with it, for backwards compatibility reasons.

urykhy:
> for backwards compatibility reasons.
i think my MR does not break compatibility. old mode in use by default, and crc32 mode must be enabled in configuration.

PR #256, opened 2020-11-29
Fix building for plan9. Fixes #255

### LINKED ISSUE #255: build fail for plan9
```
error.go:41:10: undefined: syscall.ENOTEMPTY
```

v2.1.1 - fails
v1.1.4 - ok.
build log: https://github.com/rclone/rclone/pull/4640/checks?check_run_id=1460521596

ivandeex:
This is a quick fix for the Plan9 build issue desribed in https://github.com/rclone/rclone/pull/4640#issuecomment-734450413

Reproduce by:
```
cd hdfs
GOOS=plan9 GOARCH=amd64 go build -o hdfs.plan9 -v ./cmd/hdfs
```

The issue prevents https://github.com/rclone/rclone from linking https://github.com/colinmarc/hdfs

This PR also fixes https://github.com/colinmarc/hdfs/issues/255
Please accept.

### DIFF
--- a/error_plan9.go
+++ b/error_plan9.go
@@ -1,0 +1,10 @@
+// +build plan9
+
+package hdfs
+
+import (
+	"errors"
+)
+
+// plan9 does not have ENOTEMPTY
+var osErrNotEmpty = errors.New(pathIsNotEmptyDirException)
--- a/error.go
+++ b/error.go
@@ -2,7 +2,6 @@ package hdfs
 
 import (
 	"os"
-	"syscall"
 )
 
 const (
@@ -38,7 +37,7 @@ func interpretException(err error) error {
 	case permissionDeniedException:
 		return os.ErrPermission
 	case pathIsNotEmptyDirException:
-		return syscall.ENOTEMPTY
+		return osErrNotEmpty
 	case fileAlreadyExistsException:
 		return os.ErrExist
 	default:
--- a/error_other.go
+++ b/error_other.go
@@ -1,0 +1,9 @@
+// +build !plan9
+
+package hdfs
+
+import (
+	"syscall"
+)
+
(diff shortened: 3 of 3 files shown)


error_plan9.go
```diff
@@ -0,0 +1,10 @@
+// +build plan9
+
+package hdfs
+
+import (
+	"errors"
+)
+
+// plan9 does not have ENOTEMPTY
+var osErrNotEmpty = errors.New(pathIsNotEmptyDirException)
```
colinmarc, line 10: This swallows the error. You can just reimplement the switch statement in this file.

colinmarc:
Hi @ivandeex!

I didn't know you could connect to HDFS on plan 9. Can you talk a bit about the context for this change? I'm a bit sketchy on adding (admittedly a very small amount of) code complexity for a case that doesn't really affect users of this package.

error_plan9.go
ivandeex: Yes. It's fine with me. I could not find an equivalent for plan9 so I assume it is not diagnosed on the OS level.

ivandeex:
The intention here is to let the project build and pass basic tests on _all_ supported architectures - this is prerequirement for accepting new file systems into project.
In the same time the full integration test is run only on _mainstream_ systems like windows, linux or bsd excluding more esoteric ones like plan9.

colinmarc:
Ok, but that has nothing to do with this project, I believe? I'm sorry to force you to find another solution, but I don't think it makes sense to change this project to build on plan9 when I have no plans to support that architecture.

ivandeex:
@colinmarc
Fair enough. We'll try another way - https://github.com/rclone/rclone/pull/4640#issuecomment-751737860

PR #260, opened 2021-01-14
Fix problems with SASL and Kerberos

jlisickigog:
Hello,

PR solves two problems we encountered when trying to use this tool to interact with our cluster. We have HortonWorks with Kerberos.
- The first problem was "unexpected sequence number" when trying to connect to NameNode. We have now solved it by selecting the highest possible QOP during SASL negotiation, but I cannot find a convincing explanation on the web that the highest should indeed be selected. It does not seem logical, whereas for some reason it works correctly.
- The second problem occurred when trying to upload a file, where we received an "unexpected EOF" error instead of success and an empty file on the cluster. Similarly, the file download failed. This behaviour is caused by improper SASL negotiation when connecting to a DataNode running on a privileged port. According to the https://hadoop.apache.org/docs/r3.1.0/hadoop-project-dist/hadoop-hdfs/hdfs-default.xml documentation, the `dfs.data.transfer.protection` parameter should be ignored when DataNode is listening on privileged port, which should cause the connection negotiation process to not send a SASL Handshake. 
Here  is the link [to a code](https://github.com/apache/hadoop/blob/2b9a8c1d3a2caf1e733d57f346af3ff0d5ba529c/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/protocol/datatransfer/sasl/SaslDataTransferServer.java#L108) which i believe is used by HDFS server.

### DIFF
--- a/internal/sasl/challenge.go
+++ b/internal/sasl/challenge.go
@@ -65,3 +65,16 @@ func ParseChallenge(challenge []byte) (*Challenge, error) {
 
 	return &ch, nil
 }
+
+// HighestQopLevel extracts most secure Qop level from the list provided as an argument.
+func HighestQopLevel(qopList []string) string {
+	// Search provided list for most secure qop level.
+	for _, r := range []string{QopPrivacy, QopIntegrity, QopAuthentication} {
+		for _, qop := range qopList {
+			if qop == r {
+				return r
+			}
+		}
+	}
+	return ""
+}
--- a/internal/transfer/sasl_dialer.go
+++ b/internal/transfer/sasl_dialer.go
@@ -26,10 +26,11 @@ const (
 // data protection level is specified by the server, whether it be wire
 // encryption or integrity checks.
 type SaslDialer struct {
-	DialFunc   func(ctx context.Context, network, addr string) (net.Conn, error)
-	Key        *hdfs.DataEncryptionKeyProto
-	Token      *hadoop.TokenProto
-	EnforceQop string
+	DialFunc                        func(ctx context.Context, network, addr string) (net.Conn, error)
+	Key                             *hdfs.DataEncryptionKeyProto
+	Token                           *hadoop.TokenProto
+	EnforceQop                      string
+	SkipSaslForRemotePrivilegedPort bool
 }
 
 func (d *SaslDialer) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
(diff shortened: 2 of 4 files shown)


colinmarc:
Hi @jlisickigog, thanks for the report. This sheds some light on some of the issues people have encountered. What version of HDP are you running, out of curiosity?

internal/transfer/sasl_dialer.go
```diff
@@ -49,6 +50,17 @@ func (d *SaslDialer) DialContext(ctx context.Context, network, addr string) (net
 // then returns a wrapped connection or any error encountered. In the case of
 // a protection setting of 'authentication', the bare connection is returned.
 func (d *SaslDialer) wrapDatanodeConn(conn net.Conn) (net.Conn, error) {
+	if d.SkipSaslForRemotePrivilegedPort {
```
colinmarc, line 53: Rather than making this configurable, let's just match the behavior of the java code (however objectionable) and leave a long comment.

internal/sasl/challenge.go
```diff
@@ -65,3 +65,16 @@ func ParseChallenge(challenge []byte) (*Challenge, error) {
 
 	return &ch, nil
 }
+
+// HighestQopLevel extracts most secure Qop level from the list provided as an argument.
+func HighestQopLevel(qopList []string) string {
```
colinmarc, line 70: Instead of this function, maybe change the returned challenge to have sorted Qops?

jlisickigog:
> Hi @jlisickigog, thanks for the report. This sheds some light on some of the issues people have encountered. What version of HDP are you running, out of curiosity?

HDFS 3.1.1
HDP 3.1.4.0

internal/transfer/sasl_dialer.go
jlisickigog: Reproducing the behavior of the java client requires explicit separation for Hadoop  `dfs.encrypt.data.transfer` and `dfs.data.transfer.protection` options in the client options.  When `dfs.encrypt.data.transfer` is set the privileged port feature is disabled and `dfs.data.transfer.protection` setting is ignored.  I can also imagine that the calculation of the Qop level could be delegated to a special strategy that calculates this on the fly taking into account the destination host address. This probably can also cover calculating `HighestQopLevel`. So in the end, the strategy will expose several methods for the different places where QOP level negotiation takes place.

internal/sasl/challenge.go
sgouache: Hi! Thanks for this great implementation!
I am in exactly the same configuration (HDFS 3.1.1, HDP 3.1.4.0) but have to run on other clusters as well and would love to have this modification merged upstream.
Could you please explain how you would want this to be implemented? I'd happily try to contribute a patch.
BR

colinmarc:
Merged in 6d6efa3136113bc8de10fca3065aac55542197dc and 1daceb85bf68673fc711764965a0f786e209efe4. Sorry for leaving this so long!

PR #262, opened 2021-03-02
Add a feature to sort QOP values for choosing highest priority

kimtree:
Hi, This pr is quite similar to this https://github.com/colinmarc/hdfs/pull/260, but I uploaded a new one since the author of the previous PR has no feedback yet.

The main reason that I added this feature is the same reason from the previous PR description.
> The first problem was "unexpected sequence number" when trying to connect to NameNode. We have now solved it by selecting the highest possible QOP during SASL negotiation, but I cannot find a convincing explanation on the web that the highest should indeed be selected. It does not seem logical, whereas for some reason it works correctly.

Since we need this PR to be merged ASAP, I'll reply as quickly as possible. Thanks.

### DIFF
--- a/internal/rpc/kerberos.go
+++ b/internal/rpc/kerberos.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"sort"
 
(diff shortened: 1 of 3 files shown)


colinmarc:
Merged in 6d6efa3136113bc8de10fca3065aac55542197dc.

Issue #264, opened 2021-05-17
some cmd occured error but not exit with code 1


lj99boy:
i write a sh like this

`./hdfs_cli put "$srcPath" "$remotePath"
  if [ $? -eq 0 ];then
`

but in some case , exit 0 and error occured , and run into inappropriate if

i saw code writing like this:
`
	err = filepath.Walk(source, func(p string, fi os.FileInfo, err error) error {
		if err != nil {
			fmt.Fprintln(os.Stderr, err)
			return nil
		}

		rel, err := filepath.Rel(source, p)
		if err != nil {
			fmt.Fprintln(os.Stderr, err)
			return nil
		}

		fullDest := path.Join(dest, rel)
		if fi.IsDir() {
			client.Mkdir(fullDest, mode)
		} else {
			writer, err := client.Create(fullDest)
			if err != nil {
				fmt.Fprintln(os.Stderr, err)
				return nil
			}

			defer writer.Close()
			reader, err := os.Open(p)
			if err != nil {
				fmt.Fprintln(os.Stderr, err)
				return nil
			}

			defer reader.Close()
			_, err = io.Copy(writer, reader)
			if err != nil {
				fmt.Fprintln(os.Stderr, err)
			}
		}

		return nil
	})
`

colinmarc:
I think I fixed this in a0dd10389229d32dc08a9d14e6c958283c5efaf3, although it's hard to test for. Let me know if you see it again!

PR #265, opened 2021-05-17
putting file to remote safely(add error handle)

lj99boy:
rel to https://github.com/colinmarc/hdfs/issues/264


colinmarc:
Hi @lj99boy, I don't see why the `defer` statements shouldn't work here. Can you be a bit more specific about what's happening?

I'm going to close this, but feel free to chime in with more information.

PR #271, opened 2021-08-13
Fix bug of closing file

sunzhaochang:
Fix bug of closing file.

Sometimes although the Close() method returns success, but the remote file is not really closed. Because the Result of CompleteResponseProto is not checked here.


colinmarc:
Thanks!

PR #272, opened 2021-08-19
Allow to use multiple nodes in HADOOP_NAMENODE

dotzero:
First of all, I'm happy to use your awesome library. 
But there was impossible to pass multiple nodes addresses in cli mode like this way:
```
HADOOP_NAMENODE="spark01.lan:8020,spark02.lan:8020" HADOOP_USER_NAME=dev go run ./cmd/hdfs ls /user/airflow/
```

### DIFF
--- a/cmd/hdfs/main.go
+++ b/cmd/hdfs/main.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"os/user"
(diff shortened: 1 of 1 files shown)


colinmarc:
Thanks, I can't see a reason why not!

PR #278, opened 2021-10-18
reduce allocations for header write

### LINKED ISSUE #279: Refactoring
Please check @colinmarc 
https://github.com/colinmarc/hdfs/pull/278 reduce allocations for header write
https://github.com/colinmarc/hdfs/pull/277 reduce allocation for outboundPacket
https://github.com/colinmarc/hdfs/pull/276 use only seqno instead of full outboundPacket in channel

moredure:
Fixes #279

### DIFF
--- a/internal/transfer/block_write_stream.go
+++ b/internal/transfer/block_write_stream.go
@@ -64,6 +64,7 @@ type blockWriteStream struct {
 
 	heartbeats chan struct{}
 	writeLock  sync.Mutex
+	header     bytes.Buffer
 }
 
 type outboundPacket struct {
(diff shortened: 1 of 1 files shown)


colinmarc:
Hi @moredure,

I appreciate the contribution! I can see what you're getting at, but I think it's pretty marginal for the added code complexity. Is this based on a benchmark you did, or do you have any evidence that it was a pain point for the current implementation?

I'm going to close this, but feel free to chime in with answers to the above if you think I should reconsider.

moredure:
Hi @colinmarc, this allows me to totally get rid of allocations burden per single write operations.

PR #281, opened 2021-11-10
Transparent encryption

### LINKED ISSUE #280: Transparent encryption
What do you think about [Transparent Encryption](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/TransparentEncryption.html) ? Maybe someone already working on it, or you interested in PR ?

In rclone we have https://github.com/rclone/rclone/issues/5726 to support one.

urykhy:
Fixes #280


colinmarc:
Hi @urykhy, thanks for the huge contribution!

I am interested in supporting this feature, but I lack the context to understand if the implementation is correct. If you're willing to put a bit more work in, I think we could get this merged.

To start with, I would feel better if we could add some more tests. The tests you added are a good start, but they are far from comprehensively testing this behavior. I wonder if it would make sense to run the entire test suite again (as another job on gh actions) with `/_test` marked as an encryption zone, rather than trying to duplicate tests here and there. We also need tests for the case where we don't have the key available, to make sure our error messages and exit statuses make sense.

Makefile
```diff
@@ -13,8 +13,8 @@ RELEASE_NAME = gohdfs-$(TRAVIS_TAG)-$(ARCH)
 all: hdfs
 
 %.pb.go: $(HADOOP_HDFS_PROTOS) $(HADOOP_COMMON_PROTOS)
-	protoc --go_out='$(PROTO_MAPPING):internal/protocol/hadoop_common' -Iinternal/protocol/hadoop_common -Iinternal/protocol/hadoop_hdfs $(HADOOP_COMMON_PROTOS)
-	protoc --go_out='$(PROTO_MAPPING):internal/protocol/hadoop_hdfs' -Iinternal/protocol/hadoop_common -Iinternal/protocol/hadoop_hdfs $(HADOOP_HDFS_PROTOS)
+	for f in $(HADOOP_COMMON_PROTOS); do protoc --go_out='$(PROTO_MAPPING):internal/protocol/hadoop_common' -Iinternal/protocol/hadoop_common -Iinternal/protocol/hadoop_hdfs $$f; done
```
colinmarc, line 16: Why is this change necessary? Can you leave out any proto changes from this PR, or at least make them a separate PR?

aes_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func TestAesChunks(t *testing.T) {
+	originalText := []byte("some random plain text, nice to have it quite long")
+	key := []byte("0123456789abcdef")
+
+	// choose iv to hit counter overflow
+	iv := []byte("\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\xff\xff\xff\xff\xff\xf5")
+	enc := &TransparentEncryptionInfo{iv: iv, key: key}
+
+	// one step encrypt/decrypt
```
colinmarc, line 17: If you look through the comments in the repo, you'll see I prefer full sentences with thorough explanations. Sorry for being picky. I can update some of them before merging.

Makefile
colinmarc: Also, it's weird that `make` is trying to regenerate the protos for you. They should be up-to-date in master.

Try this: 
```
find -name '*.pb.go' -exec touch {} \; && make
```

urykhy: > Maybe check the plugin has also been updated?

Yes, thanx.

aes_test.go
urykhy: Rebased with better comments.

.github/scripts/install-hdfs.sh
```diff
@@ -183,5 +221,10 @@ sleep 5
 echo "Waiting for cluster to exit safe mode..."
 $HADOOP_ROOT/bin/hdfs dfsadmin -safemode wait
 
+echo "Prepare encrypted zone"
+$HADOOP_ROOT/bin/hadoop key create key1
```
colinmarc, line 225: Mind putting this in `fixtures.sh`?

aes.go
```diff
@@ -0,0 +1,64 @@
...
+// aesCtrStep perform AES-CTR XOR operation on given byte string.
+// Once encryption and decryption are exactly the same operation for CTR mode,
+// this function can be used to perform both.
+func aesCtrStep(offset int64, enc *TransparentEncryptionInfo, b []byte) ([]byte, error) {
+	iv, err := calculateIV(offset, enc.iv)
+	if err != nil {
+		return nil, err
+	}
+
+	block, err := aes.NewCipher(enc.key)
```
colinmarc, line 49: I think we want to have this owned by the writer or reader and reuse it (or maybe the block reader?), since it [involves an allocation](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/crypto/aes/cipher.go).

kms.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	req, err := http.NewRequest("OPTIONS", url, nil)
+	if err != nil {
+		return err
+	}
+
+	var resp *http.Response
+	if c.options.KerberosClient != nil {
+		kHttp := spnego.NewClient(c.options.KerberosClient, c.http, "")
```
colinmarc, line 38: Let's put the kms stuff into a `kmsClient` and then we can reuse stuff like this.

kms.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("bad kms auth status: %v", resp.StatusCode)
+	}
+	return nil
+}
+
+func kmsSplit(keyProvider string) []string {
```
colinmarc, line 54: no need for a separate function here

kms.go
```diff
@@ -0,0 +1,186 @@
...
+	uri := defaults.GetKeyProviderUri()
+	if uri == "" {
+		return nil, errors.New("KeyProviderUri not configured on server")
+	}
+
+	var urls []string
+	var proto string
+	if strings.HasPrefix(uri, kmsSchemeHTTPS) {
+		proto = "https://"
+		urls = kmsSplit(uri[len(kmsSchemeHTTPS):])
```
urykhy, line 74: > The hostnames are separated by semi-colons, and all KMS instances must run on the same port.
via: https://hadoop.apache.org/docs/stable/hadoop-kms/index.html

but `kmsSplit` actually split by `,`. Port number must be carefully handled too.

kms.go
urykhy: i've improved kms address parsing, and this function got replaced by `kmsParseProviderUri`.

kms.go
urykhy: fixed.

file_writer.go
```diff
@@ -176,6 +201,28 @@ func (f *FileWriter) SetDeadline(t time.Time) error {
...
+	// Encrypt data chunk if file in HDFS encrypted zone.
+	if f.enc != nil && len(b) > 0 {
+		var offset int
+		for offset < len(b) {
+			size := min(len(b)-offset, aesChunkSize)
+			ciphertext, err := aesCtrStep(f.offset, f.enc, b[offset:offset+size])
+			if err != nil {
+				return offset, err
+			}
+			writtenSize, err := f.writeImpl(ciphertext)
```
colinmarc, line 213: What happens if a aes chunk gets split across multiple blocks? I guess it's fine?

We could test this by setting blocksize to 16kb with `CreateFile` and calling `Write` with 1mb of data.

aes.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	stream := cipher.NewCTR(block, iv)
+
+	padding := offset % aes.BlockSize
+	if padding > 0 {
+		tmp := make([]byte, padding)
+		stream.XORKeyStream(tmp, tmp)
+	}
+
+	text := make([]byte, len(b))
```
colinmarc, line 61: I think you can just `XORKeyStream(b, b)`

file_writer.go
urykhy: I've add test `TestEncryptedZoneLargeBlock` with hdfs block 128Kb and 1.5Mb of data. It passed.

aes.go
urykhy: no. test fail then.

aes.go
urykhy: cipher stored in `TransparentEncryptionInfo` now.

kms.go
urykhy: You mean to create `KmsClient` class, with all stuff for KMS interactions ?

.github/scripts/install-hdfs.sh
urykhy: done

file_reader.go
```diff
@@ -29,6 +30,16 @@ type FileReader struct {
 	readdirLast string
 
 	closed bool
+
+	// encryption
+	enc *TransparentEncryptionInfo
+}
+
+// A TransparentEncryptionInfo is a key and iv to encrypt or decrypt file data
+type TransparentEncryptionInfo struct {
```
colinmarc, line 39: no need to export this.

file_writer.go
```diff
@@ -176,6 +201,28 @@ func (f *FileWriter) SetDeadline(t time.Time) error {
...
+			offset += writtenSize
```
colinmarc, line 214: I think we're violating the contract of `io.Writer` by returning the output size, not the amount of bites of the input successfully written. I could be misunderstanding, though.

file_reader.go
urykhy: ok.

file_writer.go
urykhy: with encryption, we return `offset`, and `offset` updated only by `f.writeImpl` output.

file_reader.go
```diff
@@ -184,6 +209,25 @@ func (f *FileReader) Read(b []byte) (int, error) {
 		return 0, io.ErrClosedPipe
 	}
 
+	offset := f.offset
+	n, err := f.readImpl(b)
```
colinmarc, line 213: This reads the encrypted bytes into b. If the unencrypted text is smaller than the encrypted text, we'll have leftover garbage in `b`. We should (re)use a `bytes.Buffer` here if doing decryption.

file_reader.go
```diff
@@ -184,6 +209,25 @@ func (f *FileReader) Read(b []byte) (int, error) {
...
+
+	// Decrypt data chunk if file from HDFS encrypted zone.
+	if f.enc != nil && n > 0 {
+		plaintext, err := aesCtrStep(offset, f.enc, b[:n])
```
colinmarc, line 217: We should find a way to re-use the `cipher.Block` for performance.

file_reader.go
```diff
@@ -184,6 +209,25 @@ func (f *FileReader) Read(b []byte) (int, error) {
...
+		if err != nil {
+			f.offset = offset
+			return 0, err
+		}
+		for i := 0; i < n; i++ {
```
colinmarc, line 222: This can just be a `copy`.

file_reader.go
```diff
@@ -184,6 +209,25 @@ func (f *FileReader) Read(b []byte) (int, error) {
...
+			b[i] = plaintext[i]
+		}
+	}
+
+	return n, err
```
colinmarc, line 227: `n` is now the length of the ciphertext, not the plaintext. It's important to return the amount of plaintext we wrote into `b`. It's weird that this doesn't cause test failures - I would try to reproduce a pathological case in a test before fixing the bug.

file_writer.go
```diff
@@ -77,11 +82,21 @@ func (c *Client) CreateFile(name string, replication int, blockSize int64, perm
 		return nil, &os.PathError{"create", name, interpretCreateException(err)}
 	}
 
+	var enc *transparentEncryptionInfo
+	if createResp.GetFs().GetFileEncryptionInfo() != nil {
+		enc, err = c.kmsGetKey(createResp.GetFs().GetFileEncryptionInfo())
+		if err != nil {
+			c.Remove(name)
```
colinmarc, line 89: This doesn't strike me as 100% right (especially ignoring the error). Maybe we can get the key before creating the file? What does the java version do?

file_writer.go
colinmarc: Exactly. io.Writer says:

> Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early.

If the length of the written ciphertext is longer than the plaintext, then we return n > len(p). If it's less, then we return n < len(p) without an error.

My rudimentary understanding of block ciphers (confirmed by [this SO answer](https://stackoverflow.com/a/3717552) is that the ciphertext is usually, but not always, the same length as the plaintext. My guess is that the tests are passing because of an alignment between our block size and the input, but it should be possible to construct a pathological case that breaks that assumption.

file_writer.go
colinmarc: That test is helpful, but it actually just tests that our write and read code expects the same format. Ideally we would write with go and read with java to actually test that behavior.

file_writer.go
colinmarc: Btw, it's not just the interface contract that could break here: if `writtenSize` is larger than `size`, we will skip over `writtenSize - size` bytes in `b`.

file_writer_test.go
```diff
@@ -531,3 +531,83 @@ func TestFileAppendDeadlineBefore(t *testing.T) {
...
+func TestEncryptedZoneWriteChunks(t *testing.T) {
+	skipWithoutEncryptedZone(t)
+
+	originalText := []byte("some random plain text, nice to have it quite long")
+	client := getClient(t)
+	writer, err := client.Create("/_test/kms/write_chunks.txt")
+	require.NoError(t, err)
+
+	var pos int64 = 0
+	for _, x := range []int{5, 7, 6, 4, 28} {
```
colinmarc, line 550: This is a good idea. But we're still not testing the loop in `Write` or `Read`, because our data is shorter than one chunk.

file_writer_test.go
```diff
@@ -531,3 +531,83 @@ func TestFileAppendDeadlineBefore(t *testing.T) {
...
+	bytes, err := ioutil.ReadAll(reader)
+	require.NoError(t, err)
+	assert.Equal(t, originalText, bytes)
+}
+
+func TestEncryptedZoneLargeBlock(t *testing.T) {
+	skipWithoutEncryptedZone(t)
+
+	// Generate quite large (aesChunkSize * 1.5 bytes) block, so we can trigger encryption in chunks.
+	str := "some random text"
```
colinmarc, line 595: You can use `mobydick.txt` for this.

file_reader.go
colinmarc: See my comment below about block ciphers and assuming that len(in) ==len(out). Even if that is true in all the cases we can test, we shouldn't assume it's the case.

colinmarc (changes requested):

aes.go
colinmarc: Can we use `cipher.StreamReader` here?

file_reader.go
```diff
@@ -38,11 +49,25 @@ func (c *Client) Open(name string) (*FileReader, error) {
...
+			return nil, &os.PathError{"open", name, err}
+		}
+	}
+
 	return &FileReader{
 		client: c,
 		name:   name,
 		info:   info,
 		closed: false,
+		enc:    enc,
```
colinmarc, line 70: Looking at the `cipher` api a bit more, it seems to me there's a way cleaner to do this.

 - set `f.cipher` to a `cipher.Stream` with the key
 - in `Read`, deep in the loop, do 

```go
if (f.cipher != nil) { 
    n, err = cipher.StreamReader{S: f.cipher, R: f.blockReader}.Read(b) 
} else { 
    n, err = f.blockReader.Read(b)
}
```

That should be all that's needed.

file_writer.go
```diff
@@ -187,6 +234,7 @@ func (f *FileWriter) Write(b []byte) (int, error) {
 	for off < len(b) {
 		n, err := f.blockWriter.Write(b[off:])
```
colinmarc, line 235: Similarly, we could just do

```
if f.cipher != nil {
    n, err = cipher.StreamWriter{S: f.cipher, W: f.blockWriter}.Write(b[off:])
} else {
    n, err = f.blockWriter.Write(b[off:])
}
```

colinmarc:
Thanks for sticking with this!

file_writer.go
urykhy: Sure! I've add checks via 'hadoop dfs' too.

aes.go
urykhy: yes.

file_writer_test.go
urykhy: ok

file_writer.go
urykhy: It seems a bit more complicated in reality, but it's defenitely better than previous version.

file_writer.go
urykhy: as i understand, java version does not remove file:

after [dfs.create](https://github.com/naver/hadoop/blob/master/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DistributedFileSystem.java#L404=), they wrap stream with [createWrappedOutputStream](https://github.com/naver/hadoop/blob/9e1123d0a983297d48ec53f786731e52c498d6b4/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DFSClient.java#L1489=), and i see no `catch`'s to handle error.

file_writer_test.go
urykhy: After switch to `cipher.Stream` - there is no loops.

file_writer_test.go
urykhy: ok

file_reader.go
urykhy: this must not happen: encrypted text size for `ctr` mode must be the same as unencrypted.

file_reader.go
urykhy: `cipher.block` and `cipher.stream` cached now.

file_reader.go
urykhy: fixed with migration to `cipher.Stream`.

file_reader.go
urykhy: aes-ctr (the only mode supported) is a stream cipher, not block one.

Yash9060:
@colinmarc Any plans to merge this pr ?

urykhy:
@colinmarc hello, any chances ? ;)

PR #287, opened 2022-03-21
add client CopyToRemoteWithTimeout

ViYine:
when I use  CopyToRemote to upload file，sometimes the application will be blocked，
so  I add client func CopyToRemoteWithTimeout with a timeout argument  to avoid write to remote datanode blocked.

### DIFF
--- a/client.go
+++ b/client.go
@@ -307,6 +307,35 @@ func (c *Client) CopyToRemote(src string, dst string) error {
 	return remote.Close()
 }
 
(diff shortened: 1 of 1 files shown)


colinmarc:
Hi @ViYine, thanks for the PR! As this is easily achievable by just not using the convenience method, I'm going to choose not to expand the interface here. Please let me know if I'm misunderstanding something.

PR #290, opened 2022-04-09
added a count command

gardenia:
This adds basic support for the count command:

https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/FileSystemShell.html#count

Currently no command-line options supported but I find this adds value to my own use-cases as is.


colinmarc:
Hi @gardenia, thanks for the contribution. This command feels pretty un-unixy, which makes me think it's maybe not a good fit for this binary, but I'm open to discussing it. Are there usecases not covered by `ls [-R] | wc`?

Issue #299, opened 2022-08-26
AddBlockRequestProto Not Set FileId Cause FileWriter.Write Error


getumen:
error message
```
addBlock call failed with ERROR_APPLICATION (org.apache.hadoop.hdfs.server.namenode.LeaseExpiredException)
```
In Hadoop 3.x, when I uploaded a file larger than block size, LeaseExpiredException occured.
That is because AddBlockRequestProto doesn't have FileID in `addBlock` request. https://github.com/colinmarc/hdfs/blob/2f114063eda00d5847ff83c07afb9ad04e90f7b4/file_writer.go#L266-L270

I don't have, myself, full confidence that this error is reproducible in other hadoop environment.

getumen:
This issue (https://github.com/colinmarc/hdfs/issues/298) may be same reason.

982945902:
you can look at https://github.com/colinmarc/hdfs/pull/301

colinmarc:
Fixed in #301.

Issue #298, opened 2022-08-09
CompleteRequestProto Not Set FileId Cause FileWriter.Close Error


982945902:
that is failed stack:
![截屏2022-08-09 下午8 02 56](https://user-images.githubusercontent.com/49055103/183642734-f26bacf0-5e1c-4d62-b9e0-ce74d444182d.png)

under the circumstances，use "hdfs dfs -ls" show file size is zero，but use "hdfs dfs -head" can normally display file content, and use "hdfs dfs copyToLoacl" ,the downloaded file is  normally

i refer to ‘hdfs dfs’ call sequence , before call 'complete', i call 'getFileInfo' to get 'FileId', and set to CompleteRequestProto, like this:
```
func (f *FileWriter) Close() error {
	var lastBlock *hdfs.ExtendedBlockProto
	if f.blockWriter != nil {
		lastBlock = f.blockWriter.Block.GetB()

		// Close the blockWriter, flushing any buffered packets.
		err := f.finalizeBlock()
		if err != nil {
			return err
		}
	}

	var err error
	getFileInfoReq := &hdfs.GetFileInfoRequestProto{Src: proto.String(f.name)}
	getFileInfoResp := &hdfs.GetFileInfoResponseProto{}

	err = f.client.namenode.Execute("getFileInfo", getFileInfoReq, getFileInfoResp)
	if err != nil {
		return &os.PathError{"getFileInfo", f.name, err}
	}

	completeReq := &hdfs.CompleteRequestProto{
		Src:        proto.String(f.name),
		ClientName: proto.String(f.client.namenode.ClientName),
		Last:       lastBlock,
		FileId:     getFileInfoResp.Fs.FileId,
	}
	completeResp := &hdfs.CompleteResponseProto{}

	err = f.client.namenode.Execute("complete", completeReq, completeResp)
	if err != nil {
		return &os.PathError{"create", f.name, err}
	} else if completeResp.GetResult() == false {
		return &os.PathError{"create", f.name, ErrReplicating}
	}

	return nil
}
```
after changed code, i try again, successfully fix 'FileWriter.Close Error'

because of i do not known much about hdfs,  please check the cause of the problem.

982945902:
```
under the circumstances，use "hdfs dfs -ls" show file size is zero，but use "hdfs dfs -head" can normally display file content, and use "hdfs dfs copyToLoacl" ,the downloaded file is normally
```
[...]

PR #300, opened 2022-08-28
Set FileId to some RPC requests

y-tag:
We found that some RPC requests should have the FileId in our environment, which is based on the Apache Hadoop 3.3.
Thus, this PR explicitly sets FileId to the following requests.

- AddBlockRequestProto
- CompleteRequestProto


y-tag:
This may solve #298 and #299

colinmarc:
Fixed in #301. Thanks for the contribution!

PR #301, opened 2022-08-29
fix loss fileID cause write failed 

982945902:
details bind with https://github.com/colinmarc/hdfs/issues/298


colinmarc:
Thanks for the fix, @982945902!

PR #308, opened 2022-12-07
Make two source files no longer executable

mcv21:
When packaging this for Debian, the Debian linter "lintian" noticed that these two source files are executable (which they shouldn't be).

This change just removes the executable bit from the two offending source files.

Thanks, and I hope you don't mind a very pernickety PR!


colinmarc:
Huh, weird! Thanks for the fix.

PR #315, opened 2023-04-27
Change default checksum type to CRC32C

symious:
Changing the default checksum type to CRC32.

We have some users using uploading files which has the checksum type of "CRC32", but the default type is "CRC32C" since the very begining. 

This will cause a checksum mismatch issue when users are using distcp to copy the files. Although we can work around this issue on HDFS side by adding "-pc" to the distcp command, we think it's also good to change the default value from GO side.

### DIFF
--- a/internal/protocol/hadoop_hdfs/hdfs.pb.go
+++ b/internal/protocol/hadoop_hdfs/hdfs.pb.go
@@ -3609,7 +3609,7 @@ type FsServerDefaultsProto struct {
 const (
 	Default_FsServerDefaultsProto_EncryptDataTransfer = bool(false)
 	Default_FsServerDefaultsProto_TrashInterval       = uint64(0)
-	Default_FsServerDefaultsProto_ChecksumType        = ChecksumTypeProto_CHECKSUM_CRC32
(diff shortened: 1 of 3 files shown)


symious:
@colinmarc Could you help to review the PR?

mumuhhh:
You should modify the method 'writeBlockWriteRequest' in 'block_writer.go'

colinmarc:
Hi @symious,

The default checksum per the source is `CRC32`: https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs-client/src/main/proto/hdfs.proto#L530

I'm not inclined to change it here (the protos have always been a straight copy from the source), but maybe I'm missing something? I'll close this for now, but please feel free to make the argument!

symious:
@colinmarc Thank you for the reply.

The default value Clients are using should be here: https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/client/HdfsClientConfigKeys.java#L133

And CRC32C has been more efficient than the CRC32.

symious:
@colinmarc Thank you for the change.

PR #324, opened 2023-07-11
Fix skip offset of file reader Seek()

mtnk2005:
The offset of Skip() should be a offset in the block.

### DIFF
--- a/file_reader.go
+++ b/file_reader.go
@@ -163,7 +163,7 @@ func (f *FileReader) Seek(offset int64, whence int) (int64, error) {
 			// If the seek is within the next few chunks, it's much more
 			// efficient to throw away a few bytes than to reconnect and start
 			// a read at the new offset.
-			err := f.blockReader.Skip(f.offset - f.blockReader.Offset)
+			err := f.blockReader.Skip(f.offset - int64(f.blockReader.Block.GetOffset()))
 			if err == nil {
 				return f.offset, nil
 			}
--- a/file_reader_test.go
+++ b/file_reader_test.go
@@ -216,6 +216,30 @@ func TestFileReadOversizedBuffer(t *testing.T) {
 	}
 }
 
+func TestFileSeekSkip(t *testing.T) {
+	client := getClient(t)
+
+	file, err := client.Open("/_test/mobydick.txt")
+	require.NoError(t, err)
+
+	off, err := file.Seek(testStrOff, 0)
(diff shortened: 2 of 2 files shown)


colinmarc:
Hi @mtnk2005, thanks for the contribution! Can you possibly construct a pathological test case that would fail without this fix?

mtnk2005:
> Hi @mtnk2005, thanks for the contribution! Can you possibly construct a pathological test case that would fail without this fix?

ok, TestFileSeekSkip() added

colinmarc:
@mtnk2005 thanks for the quick response. I just ran the test on master, and it passed: https://github.com/colinmarc/hdfs/actions/runs/5520037428

mtnk2005:
> @mtnk2005 thanks for the quick response. I just ran the test on master (without the fix), and it passed: https://github.com/colinmarc/hdfs/actions/runs/5520037428

Hi @colinmarc, Sorry for the test, the "f.blockReader" is nil, so it didn't trigger the bug, I just fixed the test by adding a pre read(), so the "f.blockReader" != nil, and the bug should be triggered.

colinmarc:
still passing :)

https://github.com/colinmarc/hdfs/actions/runs/5520869984

I don't have an HDFS cluster to offer for realtime testing, unfortunately.

mtnk2005:
I modified and tested the code, now failed finally, please check:

[root@hadoop01 shanhuang]# ./hdfs.test -test.v -test.run TestFileSeekSkip
=== RUN   TestFileSeekSkip
2023/07/12 18:15:20 call seek
2023/07/12 18:15:20 call seek
2023/07/12 18:15:20 call skip 34 14 0
2023/07/12 18:15:20 skip 20
    file_reader_test.go:241:
        	Error Trace:	file_reader_test.go:241
        	Error:      	Not equal:
        	            	expected: "Moby"
        	            	actual  : "berg"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-Moby
        	            	+berg
        	Test:       	TestFileSeekSkip
--- FAIL: TestFileSeekSkip (0.04s)
FAIL

colinmarc:
Ahhhh - I get it now. Thank you for cracking that nut! Having a reproducible test case made it much easier for me to understand the bug. I went with a slightly different fix in 13d9368869e6f7f0a8db9d92e52c2ab54dae483e.

PR #342, opened 2025-01-22
Updated  go version and crypto dependency to fix CVE-2024-453, CVE-20…

kzgrzendek:
Hello @colinmarc :)

I tried to update the package dependencies to fix several CVE we're having on our side zith he package. It compiles and builds successfully.

The  gh workflows is not in success, but it looks like it's also the case on your master branch.

Will be happy to dicuss things further if needed ;)


colinmarc:
Hi @kzgrzendek, thanks, this looks reasonable!

